PR #18, opened 2024-03-11
Move "acknowledgements" to be part of Markdown body 

### LINKED ISSUE #12: get markdown italics to work
Right now theyre being rendered with *

rugeli:
Problem
=======
closes #12 

The issue is `acknowledgements` were not being rendered with Markdown formatting because the frontmatter stores metadata in yaml format.

Solution
========
- moved content within `acknowledgements` from frontmatter to the Markdown body
- refactored `DiseaseTable` and `DiseaseTemplate` components to query `acknowledgements` as `html` and render them using `HTMLContent`
- Tentatively adjusted css

## Type of change

Screenshot 2024-03-11 at 11 06 50 AM

### DIFF
--- a/src/style/disease-table.module.css
+++ b/src/style/disease-table.module.css
@@ -83,6 +83,7 @@
 }
 
 .container :global(.ant-table-footer p) {
+    margin: 0;
     font-size: inherit;
     color: inherit;
 }
--- a/src/style/disease-table.module.css
+++ b/src/style/disease-table.module.css
@@ -82,6 +82,10 @@
     color: var(--LIGHT_BLUE);
 }
 
+.container :global(.ant-table-footer p) {
+    font-size: inherit;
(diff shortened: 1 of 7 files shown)


src/style/disease-table.module.css
```diff
@@ -82,6 +82,11 @@
     color: var(--LIGHT_BLUE);
 }
 
+.container :global(.ant-table-footer p) {
+    margin: 0;
+    font-size: inherit;
```
meganrm, line 87: move this from line 80 (delete it there and add it here)

src/style/disease-table.module.css
```diff
@@ -82,6 +82,11 @@
...
+    color: inherit;
```
meganrm, line 88: same here with line 82

meganrm (approved):

PR #31, opened 2024-03-14
change header style

### LINKED ISSUE #25: Eliminate extra padding: There is some extra padding under the "Subscribe to our newsletter" line. Eliminate it so that the dividing line to right extends only to the bottom of the text and not beyond. Extend the Coriell button vertically so that it fills to the bottom of the dividing line (which will also be in line with the bottom of the text)
@TravisKroeker because the text on the left can wrap, the vertical space on the header is really variable. When the text is wrapped enough that it's taller than the button, do you always want the button aligned with the bottom?

meganrm:
Problem
=======
we updated the design to make the coriell button smaller 
closes #25 

Solution
========
updated the style 

## Type of change

### DIFF
--- a/src/style/disease-catalog.module.css
+++ b/src/style/disease-catalog.module.css
@@ -1,12 +1,18 @@
 .header {
-    margin-bottom: 32px;
+    margin: 16px 0 40px 0;
+}
+
+.header :global(.content) {
+    display: flex;
+    gap: 16px;
+    flex-direction: column;
 }
 
 .header p {
     font-size: 20px;
     font-weight: 400;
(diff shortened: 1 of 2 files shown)


meganrm:
@TravisKroeker, FYI, changes to the coriell button: https://deploy-preview-31--cell-catalog.netlify.app/disease-catalog/

TravisKroeker:
I'm a fan. What do you think? Send it?

rugeli (approved):

meganrm:
> I'm a fan. What do you think? Send it?

yeah I think it looks a lot better

PR #32, opened 2024-03-14
add hover effects on "obtain line" and "cert. of analysis"

### LINKED ISSUE #27: Hover effect on "obtain line" and "cert of analysis"
## Use Case

to match the prototype there should be a blue hover effect on these "buttons"

## Acceptance Criteria

when it looks like the prototype 

## Details

rugeli:
Problem
=======
closes #27

Solution
========

- modified css to apply hover effects at the cell level for the last two columns 

Note: 

- The color toggle for the three little vertical strokes within the `Tube` icon isn't working. Should we make adjustments to the css or the svg?

- I noticed that our cert svg slightly differs from the one in the design, not sure if this is significant

## Type of change

Screenshots (optional):
-----------------------
Screenshot 2024-03-14 at 2 53 34 PM

The `Cert.` icon in the design:
 
Screenshot 2024-03-14 at 2 36 01 PM

### DIFF
--- a/src/components/DiseaseTable.tsx
+++ b/src/components/DiseaseTable.tsx
@@ -17,6 +17,7 @@ const {
     clones,
     comingSoon,
     cloneNumber,
+    hoverColumn,
     footer,
 } = require("../style/disease-table.module.css");
 import { WHITE } from "./Layout";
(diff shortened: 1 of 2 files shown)


meganrm:
@rugeli I made one commit where I changed the selector to use a class name instead of the column number. This is more future proof because if we added or removed a column from the table, the selector would have gotten messed up

meganrm (approved):

rugeli:
> @rugeli I made one commit where I changed the selector to use a class name instead of the column number. This is more future proof because if we added or removed a column from the table, the selector would have gotten messed up. I also made it so the cell didn't get a hover effect if there wasn't a link

These are great changes to prevent potential issues!

PR #35, opened 2024-05-20
make disease a reference

meganrm:
time: x-small

Problem
=======
the disease had changed name but the names were hardcoded in the dropdown of the cell lines

Solution
========
changed the dropdown to a relation 

verification: 
https://deploy-preview-35--cell-catalog.netlify.app/disease-catalog/ has all three diseases, while https://main--cell-catalog.netlify.app/disease-catalog/ only has two

### DIFF
--- a/static/admin/config.yml
+++ b/static/admin/config.yml
@@ -37,12 +37,7 @@ collections:
                 widget: "number",
                 required: true,
             }
-          - {
-                label: "name",
-                name: "name",
-                widget: "string",
-                required: false,
-            }
+          - { label: "name", name: "name", widget: "string", required: false }
           - {
                 label: "Status",
                 name: "status",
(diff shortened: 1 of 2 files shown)


static/admin/config.yml
```diff
@@ -145,13 +140,10 @@ collections:
...
-                options:
-                    [
-                        "Cardiomyopathy",
-                        "Cardiomyopathy (Skeletal)",
-                        "Laminopathy",
-                    ],
+                widget: "relation",
+                collection: "disease",
+                search_fields: ["name"],
+                value_field: "name",
```
meganrm, line 146: everything in this file other than this change is just formatting

ShrimpCryptid (approved):
Side note, the Coriell Institute logo seems a little low-res and is missing alt text? Might be nice to update it if you get the chance, though that's outside the scope of this PR.

Issue #48, opened 2024-08-29
create the "editing design" data in the admin config


meganrm:
## Use Case

[the design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=1198-6040&node-type=FRAME&t=GELqnDEDzISOfI97-0) of the Editing Design subpage needs data to render

## Acceptance Criteria

the fields to add all the data needed should be visible in the admin CMS
for example: http://localhost:8000/admin/#/collections/disease-cell-line/entries/AICS-104/index

the data should all be wrapped in an object called "editing design"

there should be a list objects with the labels matching the left side of the info table in the design and the data field blank

there should be a list of "diagrams" that are objects containing a "title", an image, and a caption.  

## Details

the whole object should be collapsed on the CMS admin site by default so the data for all 6 subpages isn't overwhelming long

widget reference: https://decapcms.org/docs/widgets/#object
collapsed: if added and labeled true, collapse the widget's content by default

meganrm:
two notes: 
1. we went with "diagrams" so we should keep it consistent. 
2. we can't have default titles without then having to make things "optional" in order to submit the data, so it's better to go without a default title

PR #57, opened 2024-09-05
add editing design fields to admin config

interim17:
Problem
=======
Advances #48 

Solution
========
Added fields for the editing design to the admin config.yml
with @tyler-foster 

The only open question on this PR is how to structure the data for the `PAM sites` and `mutations` 
Substrings submitted for the fields (e.g "cRNA Target Site") will need to be formatted differently
Screenshot 2024-09-05 at 2 45 56 PM

It's not clear from the designs how the scientists will want to submit this data.
Multiple strings listed in order and tagged as default, PAM, or mutation? 
Provide one string and site/indices for PAM and mutations?

Can follow up with Jackie in Slack channel.


meganrm:
yeah that's a good question, that formatting doesn't exist on the regular cell catalog currently. I'll comment on the design for travis too, and maybe we can talk about it at wed UX hours

static/admin/config.yml
```diff
@@ -258,6 +258,69 @@ collections:
...
+                        },
+                        {
+                            label: "R primer for PCR/Sequencing",
+                            name: "r_primer",
+                            widget: "string",
+                        },
+                        {
+                            label: "Diagrams",
+                            name: "diagrams",
+                            widget: "list",
```
meganrm, line 293: I think this may need a "required: false" as well. (I haven't tested yet, but if you can't make changes to the rest of the data without having any diagrams that would be a problem)

rugeli: Yeah that's true, I have tested it. We need to add `required: false` to each field(I know)

static/admin/config.yml
```diff
@@ -258,6 +258,69 @@ collections:
...
+                            collapsed: true,
+                            fields:
+                                [
+                                    {
+                                        label: "Diagram",
+                                        name: "diagram",
+                                        widget: "object",
+                                        fields:
```
rugeli, line 301: Any thoughts on whether we need a `diagram` to wrap each set of `title`, `image` and `caption`?

interim17: Can you explain further? I see that in your PR the Diagram is a list, and ours is an object, which is how it was specified in the issue.

static/admin/config.yml
interim17: Added, thanks

static/admin/config.yml
rugeli: Yeah sure, my question was whether, under the `diagrams`(plural) list, we need a `diagram`(singular) object to contain `title`, `image` and `caption`. 

Should we follow the structure in screenshot1 or screenshot2:
Screenshot 2024-09-05 at 5 10 11 PM
Screenshot 2024-09-05 at 2 51 21 PM

Now that I'm looking at this ticket more closely, you're right - wrapping data in a `diagram` object makes more sense. 

@meganrm does screenshot1 look good to you? I can go ahead update the diagram part in our pr for consistency.

meganrm: the second one looks better to me. we should error on less nesting, both for the user, and it will make unpacking the data easier. I'm assuming that one is the list?

rugeli: Yes - the second one is a list of `Diagrams`. When `Add diagrams +` button is clicked, a new entry box is added to the list, which contains fields for `title`, `image` and `caption`, without nesting it within a `Diagram`.

static/admin/config.yml
rugeli: We need to set `required: false` for each field in `Editing Design`, like `cRNA Target Site`, `DNA Donor Sequence`, and others. This will allow us to make changes to the rest of fields without triggering any warnings.

static/admin/config.yml
interim17: Ok changed this branch to match screenshot 2 and the way your PR looks. :)

static/admin/config.yml
interim17: Ok, added to those fields.

rugeli (approved):
Could you pull the latest changes from `main` before merging? Otherwise, looks good to me!

meganrm (approved):
needs to resolve the merge issues

PR #59, opened 2024-09-12
add protocols data

### LINKED ISSUE #56: add data fields in config file for "protocols"
## Use Case

[design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=1210-7863&node-type=FRAME&t=7fKEYYsubyeLzkJ1-0)
 
## Acceptance Criteria

and optional object called "protocols" 
under than an array of files with the fields:
-file widget 
-file name 
-select of category: "cell culture/gene editing/cardiomyocyte differentiation"

## Details

rugeli:
Problem
=======
closes #56 

Solution
========
added fields for protocols data to config file 
with @meganrm @interim17 @tyler-foster 

text and format reviewed by @TravisKroeker 

## Type of change

Steps to Verify:
----------------
1. `npm run`
2. `npx netlify-cms-proxy-server`
3.  check http://localhost:8000/admin/#/collections/disease-cell-line/entries/AICS-104/index

Screenshots (optional):
-----------------------
Screenshot 2024-09-12 at 2 26 07 PM


rugeli:
Will resolve the merge conflicts once #58 is merged in `main`

interim17 (approved):
LGTM once merge changes are resolved

rugeli:
Hey Travis, based on your recent comments on the [design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=1210-7863&node-type=frame&t=o7YITDKkDzt2iiFB-0), I've updated the data entry for Protocols to include a custom category field where users can specify the header. Does the change in the screenshot match what you had in mind, or would you prefer something more explanatory? @TravisKroeker  

Screenshot 2024-09-16 at 4 01 45 PM

Also attached is the drop down menu for your reference:
Screenshot 2024-09-16 at 4 15 53 PM

TravisKroeker:
Hi Ruge,
I think this paradigm works great! Let's make a few small wording changes. For clarity, I would change the "Other" option in the dropdown menu to say "New Category" and I would change "Custom Category" to "New Category Name". 

Is there a way to make it so that the "Custom Category/New Category Name" only appears if "New Category" is selected from the dropdown? If so we should implement that. If we are able to implement it, we can eliminate the caption under "Custom Category", otherwise the reference to "other" will also have to be updated.

meganrm:
Do we not know now what these headers are going to be? If they're not really going to be changing, I think I'd just prefer we update the form once we're told what they are. I know it creates a slight bottle neck, but this seems like over complication of the data for something pretty minor. We could even just have it be called something generic like "Disease specific protocols"  in the dropdown, and then fill in the correct title on the front end depending on the disease.

tyler-foster (approved):

TravisKroeker:
I'm okay with this suggestion. Call them "Cell Culture Protocols", "Gene Editing Protocols" and "Disease Specific Protocols". There is a small chance that they don't publish Cell Culture protocols, or at least not right away.

rugeli:
Reverted to the original layout with the updated option names provided by @TravisKroeker.
Deploy preview: https://deploy-preview-59--cell-catalog.netlify.app/admin/#/collections/disease-cell-line/entries/AICS-104/index

PR #71, opened 2024-10-15
add hover styles and onlick to main disease table row when hovering o…

### LINKED ISSUE #67: implement the design for clicking to the subpage
## Use Case

a user should be able to get more information about the cell line 

## Acceptance Criteria
the hover states in[ the design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=2303-6868&node-type=frame&t=YRjxr3TFuN1Y4LMj-0) work

It only has a hover state and is clickable if the subpage data exists. (clones and images)

this shouldn't be merged until the main image work is done, otherwise it looks like an incomplete page.  

## Details

interim17:
Closes #67 

Merging blocked.

Problem
=======
This adds hover styling and `onClick` navigation to disease table rows when they have complete data.

Event propagation is blocked in `ParentalLineModal `so that modal can be opened and closed without triggering the parent row's `onClick`.

My understanding is that this should only be functional if the status of the disease line is `"data complete"`, which isn't true for the disease lines currently, so the merged changes will not seem functional. 

To see the hover states and make links functional, replace `record.status === CellLineStatus.DataComplete` on L126 and L139 with `true`

### DIFF
--- a/src/components/DiseaseTable.tsx
+++ b/src/components/DiseaseTable.tsx
@@ -47,7 +47,6 @@ const DiseaseTable = ({
 }: DiseaseTableProps) => {
     const [hoveredRowIndex, setHoveredRowIndex] = useState(-1);
     const inProgress = status?.toLowerCase() === "coming soon";
-    const CELL_LINE_PATH = `/${diseaseCellLines[0].templateKey}/AICS-`;
 
     const width = useWindowWidth();
     const isTablet = width < TABLET_BREAKPOINT;
@@ -124,7 +123,7 @@ const DiseaseTable = ({
             onMouseLeave: () => setHoveredRowIndex(-1),
             onClick: () => {
                 if (record.status === CellLineStatus.DataComplete) {
-                    navigate(CELL_LINE_PATH + record.cell_line_id);
+                    navigate(record.path);
                 }
             },
         };
@@ -166,9 +165,7 @@ const DiseaseTable = ({
                             );
                             return record.status ===
                                 CellLineStatus.DataComplete ? (
-                                <Link to={CELL_LINE_PATH + cell_line_id}>
-                                    {cellLine}
-                                </Link>
+                                <Link to={record.path}>{cellLine}</Link>
                             ) : (
                                 cellLine
                             );
--- a/src/components/DiseaseTable.tsx
+++ b/src/components/DiseaseTable.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from "react";
+import { Link } from "gatsby";
 import { Table, Tag, Flex } from "antd";
 import Icon from "@ant-design/icons";
 
@@ -119,8 +120,10 @@ const DiseaseTable = ({
             onMouseEnter: () => setHoveredRowIndex(index),
             onMouseLeave: () => setHoveredRowIndex(-1),
             onClick: () => {
-                if (record.status === CellLineStatus.DataComplete)
+                if (record.status === CellLineStatus.DataComplete) {
+                    // if (true)
                     window.location.href = `/disease-cell-line/AICS-${record.cell_line_id}`;
+                }
             },
         };
     };
@@ -153,11 +156,20 @@ const DiseaseTable = ({
                         className: cellLineId,
                         dataIndex: "cell_line_id",
                         fixed: "left",
-                        render: (cell_line_id: string) => (
-                            <h4 key={cell_line_id}>
-                                {formatCellLineId(cell_line_id)}
-                            </h4>
-                        ),
+                        render: (cell_line_id, record) =>
+                            record.status === CellLineStatus.DataComplete ? (
+                                <Link
+                                    to={`/disease-cell-line/AICS-${cell_line_id}`}
+                                >
+                                    <h4 key={cell_line_id}>
+                                        {formatCellLineId(cell_line_id)}
+                                    </h4>
+                                </Link>
+                            ) : (
+                                <h4 key={cell_line_id}>
+                                    {formatCellLineId(cell_line_id)}
+                                </h4>
+                            ),
                         onCell: onCellInteraction,
                     },
                     {
--- a/src/components/DiseaseTable.tsx
+++ b/src/components/DiseaseTable.tsx
@@ -45,6 +45,7 @@ const DiseaseTable = ({
 }: DiseaseTableProps) => {
     const [hoveredRowIndex, setHoveredRowIndex] = useState(-1);
     const inProgress = status?.toLowerCase() === "coming soon";
+    const CELL_LINE_PATH = `/${diseaseCellLines[0].templateKey}/AICS-`;
 
     const width = useWindowWidth();
     const isTablet = width < TABLET_BREAKPOINT;
@@ -121,8 +122,7 @@ const DiseaseTable = ({
             onMouseLeave: () => setHoveredRowIndex(-1),
             onClick: () => {
                 if (record.status === CellLineStatus.DataComplete) {
-                    // if (true)
-                    window.location.href = `/disease-cell-line/AICS-${record.cell_line_id}`;
+                    window.location.href = CELL_LINE_PATH + record.cell_line_id;
                 }
             },
         };
@@ -156,20 +156,21 @@ const DiseaseTable = ({
                         className: cellLineId,
                         dataIndex: "cell_line_id",
                         fixed: "left",
-                        render: (cell_line_id, record) =>
-                            record.status === CellLineStatus.DataComplete ? (
-                                <Link
-                                    to={`/disease-cell-line/AICS-${cell_line_id}`}
-                                >
-                                    <h4 key={cell_line_id}>
-                                        {formatCellLineId(cell_line_id)}
-                                    </h4>
-                                </Link>
-                            ) : (
+                        render: (cell_line_id, record) => {
+                            const cellLine = (
                                 <h4 key={cell_line_id}>
                                     {formatCellLineId(cell_line_id)}
                                 </h4>
-                            ),
+                            );
+                            return record.status ===
+                                CellLineStatus.DataComplete ? (
+                                <Link to={CELL_LINE_PATH + cell_line_id}>
+                                    {cellLine}
+                                </Link>
+                            ) : (
+                                cellLine
+                            );
+                        },
                         onCell: onCellInteraction,
                     },
                     {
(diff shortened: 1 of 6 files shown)


src/components/DiseaseTable.tsx
```diff
@@ -103,6 +105,29 @@ const DiseaseTable = ({
...
+    const handleCellInHoveredRow = (
+        record: UnpackedDiseaseCellLine,
+        index: number | undefined
+    ) => {
+        const hoveredAndDataComplete =
+            index === hoveredRowIndex && record.status === "data complete";
+        return {
+            className: hoveredAndDataComplete ? hoveredRow : "",
+            onMouseEnter: () => {
+                if (index !== undefined) {
```
meganrm, line 117: I would move this to an early return 
`if (index === undefined) {
            return {};
        }` 
because there isn't a meaningful state that you're checking for here, it's just a typescript issue

src/components/DiseaseTable.tsx
```diff
@@ -103,6 +105,29 @@ const DiseaseTable = ({
         ),
     };
 
+    const handleCellInHoveredRow = (
```
meganrm, line 108: this may be me, so feel free to disregard, but in your attempt to make this name really clear, I find it kinda hard to read. I think `handleCell` or `handleCellInteraction` or `onCell` or `onCellInteraction` would all be easier to read

src/components/DiseaseTable.tsx
```diff
@@ -103,6 +105,29 @@ const DiseaseTable = ({
...
+        record: UnpackedDiseaseCellLine,
+        index: number | undefined
+    ) => {
+        const hoveredAndDataComplete =
+            index === hoveredRowIndex && record.status === "data complete";
```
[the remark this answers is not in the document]
meganrm: we have an enum for this so you shouldn't use the hardcoded string

src/components/DiseaseTable.tsx
```diff
@@ -103,6 +105,29 @@ const DiseaseTable = ({
...
+        return {
+            className: hoveredAndDataComplete ? hoveredRow : "",
```
meganrm, line 115:
```suggestion
            className: index === hoveredRowIndex ? hoveredRow: "", 
```
and then use the rowClassName in the `Table`
```                
rowClassName={(record) =>
                    record.status === CellLineStatus.DataComplete
                        ? dataComplete
                        : ""
                }
```

The reason for this is then you can use a css selector without any antd classes. You can have `.data-complete .hovered-row` to do your styling. Also, the `data-complete` is true regardless of hovering, so it makes sense that class would be consistent, while the hover state should be dynamic.

src/components/DiseaseTable.tsx
```diff
@@ -103,6 +105,29 @@ const DiseaseTable = ({
...
+            onMouseEnter: () => {
+                if (index !== undefined) {
+                    setHoveredRowIndex(index);
+                }
+            },
+            onMouseLeave: () => {
+                setHoveredRowIndex(-1);
+            },
+            onClick: () => {
+                if (hoveredAndDataComplete)
```
meganrm, line 125: this can just be `if (record.status === CellLineStatus.DataComplete` because if you're clicking it, it's also hovered. I also wonder if this would be broken on tablets since there really isn't a hover state. It's safer to have the if statement be only for what you truly need to be true

src/components/DiseaseTable.tsx
```diff
@@ -24,6 +24,7 @@ const {
     clones,
     cellLineId,
     expandableContent,
+    hoveredRow,
```
meganrm, line 27: I think `hoverColumn` could be renamed to keep the style sheet from being confusing now that we have hoverRow. `actionColumn` seems good?

src/components/DiseaseTable.tsx
```diff
@@ -103,6 +105,29 @@ const DiseaseTable = ({
...
+                if (index !== undefined) {
+                    setHoveredRowIndex(index);
+                }
+            },
+            onMouseLeave: () => {
+                setHoveredRowIndex(-1);
+            },
+            onClick: () => {
+                if (hoveredAndDataComplete)
+                    window.location.href = `/disease-cell-line/AICS-${record.cell_line_id}`;
```
meganrm, line 126: This doesn't seem to be causing build errors, but gatsby has a navigate function 
`import { navigate } from 'gatsby';` for this kind of thing. I'm not sure if there are advantages or not

src/style/disease-table.module.css
```diff
@@ -44,6 +44,11 @@
     display: none; /* remove dividers between column headers */
 }
 
+.container :global(.ant-table-row) .hovered-row:global(.ant-table-cell) {
```
meganrm, line 47: do you need the `:global(.ant-table-cell)` here?

src/components/DiseaseTable.tsx
interim17: Haha, makes me laugh because I was trying to be extra explicit in your footsteps, thanks for feedback. I think I like `onCellInteraction` but definitely open minded here.

src/components/DiseaseTable.tsx
interim17: Ohh, thank you. I don't know why I didn't think of that. I went looking for react-router.

src/style/disease-table.module.css
interim17: I need it for specificity, working with some of your changes if I use this:
```
.container .data-complete .hovered-row:global(.ant-table-cell) 
```
it works, and if I take out the `:global(.ant-table-cell)` and add `!important` it also works, so I think
`.container .data-complete .hovered-row` targets the cells but isn't specific enough to override some other rule.

src/components/DiseaseTable.tsx
```diff
@@ -103,11 +107,34 @@ const DiseaseTable = ({
...
                 key={diseaseName}
                 className={[container, inProgress ? comingSoon : ""].join(" ")}
+                rowClassName={(record) =>
+                    record.status === CellLineStatus.DataComplete
+                        ? dataComplete
+                        : ""
+                }
                 title={() => (
                     <Flex align="center">
                         <h3 className={tableTitle}>{diseaseName}</h3>
```
meganrm, line 140:
```suggestion
                        {record.status === CellLineStatus.DataComplete? <Link to={record.cell_line_id}><h3 className={tableTitle}>{diseaseName}</h3></Link> : <h3 className={tableTitle}>{diseaseName}</h3>
```

Don't commit this directly because I'm just writing it in the GitHub UI. also you could define it up above, ie `const cellLineName = ` to keep the JSX from being too messy. But I think it would be good to have true link in here even though it's redundant with the click handler

src/style/disease-table.module.css
```diff
@@ -44,15 +44,24 @@
     display: none; /* remove dividers between column headers */
 }
 
-.container .hover-column:global(.ant-table-cell):has(a):hover {
+.container .data-complete .hovered-row:global(.ant-table-cell) {
```
meganrm, line 47: do you need `:global(.ant-table-cell)` here? it seems like 3 class names should be specific enough, and if possible it's nice to not have to use the global selector

interim17: It's specific enough for the pointer but not the background-color, I lose that style if I remove it.

src/components/DiseaseTable.tsx
interim17: I'm following your thought but I'm pretty sure you mean this to be down on 156-160 where the name of the disease cell line is printed, no where the disease name is, so I'm making the change there.

meganrm: yeah, you're right

rugeli (approved):
LGTM! I was able to see the hover effects and navigate to the subpage. Just a small note: in the design, the color of gene symbol tag button is slightly lighter than the hover background color. 
Here's the [design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=1057-5850&node-type=frame&t=btLLvQcFY3hAgwAs-0) I was looking at, not sure if it's the most recent version though.

meganrm:
> LGTM! I was able to see the hover effects and navigate to the subpage. Just a small note: in the design, the color of gene symbol tag button is slightly lighter than the hover background color.
> Here's the [design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=1057-5850&node-type=frame&t=btLLvQcFY3hAgwAs-0) I was looking at, not sure if it's the most recent version though.

this is the correct design https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=2303-6868&node-type=frame&t=RzXVX4gtBrgvGulB-0

rugeli:
> > LGTM! I was able to see the hover effects and navigate to the subpage. Just a small note: in the design, the color of gene symbol tag button is slightly lighter than the hover background color.
> > Here's the [design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=1057-5850&node-type=frame&t=btLLvQcFY3hAgwAs-0) I was looking at, not sure if it's the most recent version though.
> 
> this is the correct design https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=2303-6868&node-type=frame&t=RzXVX4gtBrgvGulB-0

Oh great, then the changes look identical to it!

src/components/DiseaseTable.tsx
```diff
@@ -126,11 +158,22 @@ const DiseaseTable = ({
...
+                            const cellLine = (
+                                <h4 key={cell_line_id}>
+                                    {formatCellLineId(cell_line_id)}
+                                </h4>
+                            );
+                            return record.status ===
+                                CellLineStatus.DataComplete ? (
+                                <Link to={CELL_LINE_PATH + cell_line_id}>
+                                    {cellLine}
+                                </Link>
```
meganrm, line 171:
```suggestion
                   <Link to={record.path}>{cellLine}</Link>

```

src/components/DiseaseTable.tsx
```diff
@@ -103,11 +111,35 @@ const DiseaseTable = ({
...
+        if (index === undefined) {
+            return {};
+        }
+        return {
+            className: index === hoveredRowIndex ? hoveredRow : "",
+            onMouseEnter: () => setHoveredRowIndex(index),
+            onMouseLeave: () => setHoveredRowIndex(-1),
+            onClick: () => {
+                if (record.status === CellLineStatus.DataComplete) {
+                    navigate(CELL_LINE_PATH + record.cell_line_id);
```
meganrm, line 127:
```suggestion
                    navigate(record.path);
```

meganrm: we already unpacked the path which was fields.slug

meganrm (approved):

PR #75, opened 2024-10-25
add and render clone data

rugeli:
Problem
=======
The clone table currently displays only the `Clone Type` column. 

Solution
========
Added the rest of clone data

## Type of change

Screenshots (optional):
-----------------------
Screenshot 2024-10-25 at 12 53 08 PM

### DIFF
--- a/src/components/CloneTable.tsx
+++ b/src/components/CloneTable.tsx
@@ -11,7 +11,7 @@ interface CloneTableProps {
 const CloneTable: React.FC<CloneTableProps> = ({ dataSource }) => {
     // NOTE: once the clone data is filled in, we can use the id as the key
     const dataSourceWithKey = dataSource.map((data, index) => {
-        return { ...data, key: index };
+        return { ...data, clone_number: `cl.${data.clone_number}`, key: index };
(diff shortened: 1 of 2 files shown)


src/components/CloneTable.tsx
```diff
@@ -11,7 +11,7 @@ interface CloneTableProps {
 const CloneTable: React.FC<CloneTableProps> = ({ dataSource }) => {
     // NOTE: once the clone data is filled in, we can use the id as the key
```
rugeli, line 12: Just checking - does `id` refer to `cell_line_id`(i.e.`cellLine.frontmatter.cell_line_id`)?

src/components/CloneTable.tsx
```diff
@@ -11,7 +11,7 @@ interface CloneTableProps {
...
     const dataSourceWithKey = dataSource.map((data, index) => {
-        return { ...data, key: index };
+        return { ...data, clone_number: `cl.${data.clone_number}`, key: index };
```
meganrm, line 14: key can also be the `clone_number` if it exists

PR #92, opened 2025-02-05
Allow multiple genes entry for normal cell lines

rugeli:
Problem
=======
The scientists want to enter multiple genes for each normal cell line.

Solution
========
- set `multiple: true` for the `gene` field in config.yml
- updated the cell line cms preview to display multiple entries, joined by commas

## Type of change

Steps to Verify:
----------------
1. Click a normal cell line and enter multiple genes in admin page

Expected results:
Screenshot 2025-02-05 at 3 06 05 PM

Screenshot 2025-02-05 at 3 00 32 PM

### DIFF
--- a/static/admin/config.yml
+++ b/static/admin/config.yml
@@ -94,6 +94,7 @@ collections:
                 value_field: "symbol",
                 display_fields: ["symbol"],
                 required: false,
+                multiple: true,
             }
           - {
                 label: "Allele Count",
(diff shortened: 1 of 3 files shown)


static/admin/config.yml
```diff
@@ -94,6 +94,7 @@ collections:
                 value_field: "symbol",
                 display_fields: ["symbol"],
                 required: false,
+                multiple: true,
```
interim17, line 97: Do any queries need to be updated to handle this?

interim17 (approved):

static/admin/config.yml
rugeli: Yes, you're right! The queries and types will need to be updated in multiple places to properly handle multiple genes(i.e., switching from a string to an array). I also noticed the tag fields allow multiple entires but are handled as strings. Since this update could impact the current UI, I'll double check with Megan and update the types for both tags and genes, possibly in a separate pr.

rugeli:
Follow up ticket: #97

PR #95, opened 2025-02-08
Feature/editing design tab

### LINKED ISSUE #93: Render "Editing Design" Tab
## Use Case

we need to visualize the data for the Editing Design

## Acceptance Criteria

it should match [the design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=1210-7863&t=0JIe0EFdlTp2BGQe-0)

1. need to request the `editing_design` data in the graphql query 
2. Send it into the Subpage component 
3. render it to match the design (including the color for the PAM site and mutation)

interim17:
Problem
=======
Closes #93 

Solution
========

Expand disease cell line query to include editing data
New `Subpage`... hopefully reusable for other subpages? Styles the cards.
`EditingDesign` to provide data to an instance of `Subpage`

Screenshots (optional):
-----------------------
Screenshot 2025-02-07 at 4 37 39 PM
Screenshot 2025-02-07 at 4 37 31 PM


src/style/disease-subpage.module.css
```diff
@@ -0,0 +1,86 @@
...
+
+.diagram-card {
+    margin-top: 24px;
+    border: 1px solid var(--ALLEN_BLUE);
+    position: relative;
+    padding-bottom: 40px;
+}
+
+.diagram-card-footer {
+    position: absolute;
```
rugeli, line 69: one small thing: multi-line captions are being cut off. I tried removing the fixed height so they can be wrapped, what do you think?

src/components/Subpage.tsx
```diff
@@ -0,0 +1,128 @@
...
+    };
+    headerLeadText?: string;
+}
+
+interface DiagramCardProps extends CardProps {
+    title?: string;
+    footer?: React.ReactNode;
+}
+
+export const DiagramCard: React.FC<DiagramCardProps> = ({
```
meganrm, line 39: I'm not sure if subpage makes sense as a separate component, but this DiagramCard should definitely be it's own component, because this will be reused a lot

src/types.ts
```diff
@@ -10,3 +12,13 @@ export enum Disease {
     Laminopathy = "Laminopathy",
     SkeletalMyopathy = "Skeletal Myopathy",
 }
+
+export interface Diagram {
```
meganrm, line 16: this seems like it should be a props interface, not a separate type out here (since it'll probably only be used where the component is)

meganrm:
Check `shared/InfoPanel` to see if that can be used for the table

src/components/Subpage.tsx
meganrm, line 1: to me 'Subpage' might make more sense as a folder, that has the tabs component, and all of the separate subpages

src/components/Subpage.tsx
```diff
@@ -0,0 +1,128 @@
...
+    <Card
+        className={className}
+        title={title && <div>{title}</div>}
+    >
+        {children}
+        {footer && <div className={diagramCardFooter}>{footer}</div>}
+    </Card>
+);
+
+const formatTextWithGeneLocations = (text: string, className: string) => {
```
meganrm, line 54: this is very specific to editing design, and should be in there (or in an associated util file)  and not in a shared component because I don't think it will be used anywhere else

src/components/SubPage/EditingDesignSubpage.tsx
```diff
@@ -0,0 +1,85 @@
...
+            );
+        }
+        return part;
+    });
+};
+
+const EditingDesignContent: React.FC<EditingDesignContentProps> = ({
+    editingDesignData,
+}) => {
+    const contentData = React.useMemo(
```
meganrm, line 40: this doesn't need a useMemo: it's the only prop, it'll re-render every time the prop changes, so you're not saving any render cycles here unless I'm missing something

src/components/SubPage/EditingDesignSubpage.tsx
```diff
@@ -0,0 +1,85 @@
+import React from "react";
+import SubpageContent from "./Subpage";
```
meganrm, line 2: nit: should be named the same as the Component (confusing otherwise)

src/components/SubPage/Subpage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    );
+
+    return (
+        <div className={container}>
+            <Card className={card} bordered={true}>
+                <div className={dataTable}>{data.rows.map(renderDataRow)}</div>
+                {data.legendContent && (
+                    <div className={legendText}>{data.legendContent}</div>
+                )}
+            </Card>
```
meganrm, line 75: the other subpages don't have data tables at the top, so this doesn't seem very general as is. If you're trying to make a reusable component here, I think what would be most helpful is  a general "SubpageContentCard" component that doesn't care what it renders (could be an image or could be an InfoPanel or could be a table) that takes in a "content", optional "title", and optional "footer", and I would put it in "shared/". Right now this is halfway between general and specific

src/components/SubPage/Subpage.tsx
```diff
@@ -0,0 +1,108 @@
...
+export interface DataRow {
+    label: string;
+    value: string | React.ReactNode;
+}
+
+export interface SubpageProps {
+    data: {
+        rows: DataRow[];
+        diagrams?: Diagram[];
+        legendContent?: React.ReactNode;
```
meganrm, line 28: for clarity, I would make these all top level props, instead of grouping them in data. that way it's clear what is being sent into the component

src/components/SubPage/Subpage.tsx
interim17: ok! for context, I haven't finished working on your other feedback I just pushed some initial commits before i left the office. maybe i should have waited to push?

for now i've just made the subfolder really, was planning to consolidate further, and I haven't really looked at `InfoCard` yet. I can wait till it's finished to push more commits if that's confusing, I'll be sure to re-request when I think it's ready.

interim17:
Ok so:
- removed `Subpage`
- created `Subpage` folder for tabs and subpages
- `DiagramCard` component now  in `/shared`
- `EditingDesignSubpage` renamed and uses `InfoPanel`

I left the `Diagram` as a more general type because it's currently used in `EditingDesignSubpageProps`, `DiagramCardProps` and for the query interface `DiseseCellLineFrontMatter`. I imagine we will also use it in the other subpages. Unless I misunderstood your feedback @meganrm ?

src/component-queries/types.ts
```diff
@@ -1,4 +1,5 @@
 import { FileNode } from "gatsby-plugin-image/dist/src/components/hooks";
+import { Diagram } from "../types";
```
meganrm, line 2: I would move this type into this file and also export it from there. All the types that are the database structure should be here, the more general `types` is for enforcing strings and such

src/components/SubPage/EditingDesignSubpage.tsx
```diff
@@ -0,0 +1,105 @@
...
+const { pamSite, mutation } = require("../../style/editing-design.module.css");
+
+interface EditingDesignSubpageProps {
+    editingDesignData: {
+        crna_target_site: string;
+        dna_donor_sequence: string;
+        cas9: string;
+        f_primer: string;
+        r_primer: string;
+        diagrams: Diagram[];
```
meganrm, line 22: I think these should just be top level props. I also think the "convert-data.ts" could be a good place to take the database data and convert it into front end consumable data. It's where I've converted everything from snake case to camel case too

src/component-queries/types.ts
interim17: Ok will do.

Also, question, i pushed the last commit the checks were failing here AND my local run had an error, after a moment my local errors resolved. 

Because my local instance went from broken to not broken without making any changes I'm assuming it's a netlify/sync issue? How do I re-run the jobs above to see?

src/style/disease-subpage.module.css
```diff
@@ -0,0 +1,82 @@
...
+    font-weight: 400;
+    color: var(--MEDIUM_GRAY);
+}
+
+.legend-text {
+    font-size: 14px;
+    font-size: 16px;
+}
+
+.diagram-card {
```
meganrm, line 50: this should all be in its own module called diagram-card to match the component

src/components/shared/DiagramCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface Diagram {
+    image: {
+        childImageSharp: {
+            gatsbyImageData: IGatsbyImageData;
+        };
+    };
+    title: string;
+    caption: string;
+}
```
meganrm, line 19: sorry, I was confused before about this interface, I didn't realize it was also the query result, so my comments probably were confusing. my only suggestion other than just importing the interface Diagram from component-queries/types.ts is you could adjust the data so that image you send into the component is just the iGatsbyImageData, since there is no need for it to be nested anymore.

src/components/SubPage/SubpageTabs.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const SubpageTabs: React.FC<SubpageTabsProps> = ({
+    tabsToRender,
+    editingDesignData,
+}) => {
+    // TODO: Swap out actual components for placeholders
+    const tabComponents = {
+        [SubPage.EditingDesign]: (
+            <EditingDesignSubpage editingDesignData={editingDesignData} />
```
meganrm, line 24:
```suggestion
            <EditingDesignSubpage {...editingDesignData} />
```
so that the props are still clean for the subpage tab but the editing design subpage doesn't have to unpack all the props from an object

meganrm: I also think this is better for react: I don't think it does diffs on nested prop objects (but I could be wrong)

src/component-queries/types.ts
```diff
@@ -145,6 +164,15 @@ export interface UnpackedNormalCellLine extends UnpackedCellLineMainInfo {
...
 
+export interface EditingDesignData {
+    crnaTargetSite?: string;
+    dnaDonorSequence?: string;
+    cas9?: string;
+    fPrimer?: string;
+    rPrimer?: string;
+    diagrams?: Diagram[];
+}
+
```
meganrm, line 175: this is redundant with EditingDesignSubpageProps, you don't need this here

interim17: This interface is used to type the editing design data when it gets unpacked, and the editing design section on the `UnpackedDiseaseCellLine` (which we aren't fully implementing yet, because there are two things both called `DiseaseCellLineTemplate` and one generally takes unpacked data and one doesn't, we should scope a ticket for that?).

If the plan is to try and unpack the data into JS objects with camel case names before sending it into our react components, this interface is useful and we should keep it.

meganrm (approved):

src/style/disease-subpage.module.css
meganrm: I'm doing this separately, so you can go ahead and merge this

src/style/disease-subpage.module.css
```diff
@@ -0,0 +1,82 @@
...
+
+.container :global(.ant-descriptions-item-label) {
+    font-weight: 400;
+    color: var(--MEDIUM_GRAY);
+}
+
+.legend-text {
+    font-size: 14px;
+    font-size: 16px;
+}
```
meganrm, line 48: setting the font size twice

PR #104, opened 2025-03-06
disable wheel hook for number inputs 

### LINKED ISSUE #98: Disable up/down arrows on number fields in CMS
## Use Case

The number fields in CMS have default up/down arrows, which allows users to increment or decrement values by scrolling. The values can be accidentally modified without users realizing it.  

## Acceptance Criteria

- Disable the arrows for number fields to prevent accidental scrolling changes.
- Users can still manually input numbers.

rugeli:
Problem
=======
closes #98 

Solution
========
- add a hook that adds an event listener to blur `number` inputs whenever a mouse wheel event occurs

with @meganrm 

## Type of change

Steps to Verify:
----------------
1. In the cms, scroll the mouse wheel while a number input is in focus

### DIFF
--- a/src/cms/hooks/useDisableWheel.ts
+++ b/src/cms/hooks/useDisableWheel.ts
@@ -1,0 +1,16 @@
+import { useEffect } from "react";
+
+const useDisableWheel = () => {
+    useEffect(() => {
+        document.addEventListener("wheel", function (event) {
+            if (
+                document.activeElement &&
+                (document.activeElement as HTMLInputElement).type === "number"
+            ) {
+                (document.activeElement as HTMLInputElement).blur();
+            }
+        });
+    }, []);
+};
+
+export default useDisableWheel;
--- a/src/cms/preview-templates/CellLinePreview.tsx
+++ b/src/cms/preview-templates/CellLinePreview.tsx
@@ -1,12 +1,16 @@
-import React from 'react'
-import { CellLineTemplate } from '../../templates/cell-line'
-import { TemplateProps } from './types';
+import React from "react";
+import { CellLineTemplate } from "../../templates/cell-line";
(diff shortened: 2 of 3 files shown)


src/cms/hooks/useDisableWheel.ts
```diff
@@ -0,0 +1,16 @@
...
+        document.addEventListener("wheel", function (event) {
+            if (
+                document.activeElement &&
+                (document.activeElement as HTMLInputElement).type === "number"
+            ) {
+                (document.activeElement as HTMLInputElement).blur();
+            }
+        });
+    }, []);
+};
```
ShrimpCryptid, line 14: Will this also blur on laptop touchpad scrolling? Is that behavior jarring if a user is in the process of editing an input field and scrolls slightly?

ShrimpCryptid: Blurring on any scroll event seems kind of heavy-handed as a solution. Is it possible to just disable the handlers on the numeric input fields? Is there a configuration option in the CMS that lets you change this?

meganrm: unfortunately no, we haven't found one. Right now this is causing more issues than the connivence of scrolling helps. Your first question sounds like the issue: the user was trying to scroll down the page and accidentally changed the number value and didn't notice

rugeli: Yeah this hook targets only the number fields in the CMS admin page, and it also works on touchpads. The goal of this approach is to prevent accidental changes to numeric values. I've checked out decap cms options for number widget, but it currently doesn't have a option to handle this.

ShrimpCryptid (approved):

meganrm (approved):

Issue #106, opened 2025-03-06
Move clone data all into the `clones` object


meganrm:
## What needs to happen?
in addition to `type`, `clone_number`, and `transfection_replicate`, 
clones should have 
1. positive_cells        
```YAML
 {
                            label: "Troponin T, % Positive Cells(n)",
                            name: "positive_cells",
                            widget: "float",
                            required: false,
}
```
 (see #105 )

2. `antibody_analysis`, multiselect list: PAX6, SOX17, Brachyury. (I can't remember if you can let the user add new ones, if so, allow that) 
```YAML
 {
                            label: "Passing Antibodies",
                            name: "antibody_analysis",
                            hint: "select all antibodies this clone passed ",
                           widget: "select",
                            options: [PAX6, SOX17, Brachyury],
                    multiple: true,

}
```
4. differentiation: list of objects, with each of the transcription factors as the key, and the value a float. (see the design)

remove `stem_cell_characteristics`
move all the `positive_cells` into these new data formats in each of the index.md files

PR #108, opened 2025-03-06
clones data structure

### LINKED ISSUE #105: add `valueType: float` in `positive_cells`
## What needs to happen?

the % positive_cells needs to be a float

interim17:
Problem
=======
Closes #105 and #106 

Solution
========

Added `positive_cells`, `antibody_analysis`, and `differentiation` to `clones`.
Deleted `stem_cell_characteristics.`

Data was updated and I found one discrepancy:
clone number 88 in AICS-106 had a value for `positive_cells` but not a clone entry. I created a clone entry, and chose type: Control, as that is a required field, but I don't know what it actually should be, or any of the other data associated with it.

Letting users add new antibodies is not possible with the basic select widget.

### DIFF
--- a/src/pages/disease-cell-line/AICS-106/index.md
+++ b/src/pages/disease-cell-line/AICS-106/index.md
@@ -13,9 +13,6 @@ clones:
     genotype: WT/WT
     positive_cells: 73
   - type: Control
-    clone_number: 88
-    positive_cells: 66
-  - type: Control
     clone_number: 86
     transfection_replicate: B
     genotype: WT/WT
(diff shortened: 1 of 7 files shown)


src/pages/disease-cell-line/AICS-106/index.md
```diff
@@ -11,6 +11,10 @@ clones:
     clone_number: 40
     transfection_replicate: A
     genotype: WT/WT
+    positive_cells: 73
+  - type: Control
+    clone_number: 88
+    positive_cells: 66
```
meganrm, line 17: I would just make a note of this in an issue and ask about it when Jacqueline gets back. it's probably a mistake

interim17: #109

meganrm (approved):

rugeli (approved):

PR #110, opened 2025-03-07
Render multiple sequences in editing design 

### LINKED ISSUE #103: switch `dna_donor_sequence` widget to `list` in `editing_design`
## What needs to happen?

- [ ] change to list
- [ ] edit exiting data to be a list

rugeli:
Problem
=======
closes #103

Solution
========
- updated `dna_donor_sequence` widget to use `list` type
- added `type` options for each sequence
- updated `EditingDesign` component to render multiple sequences   

with @meganrm 

## Type of change

## Question: how critical and unique is the asterisk (`*`) in a mutation? 
should we store the asterisk as part of the mutation text in md files, or can it be placed in a HTML, like `<span>`? 

current look for AICS-107 (not yet placed the asterisk)
Screenshot 2025-03-07 at 1 37 05 PM


src/components/SubPage/EditingDesign.tsx
```diff
@@ -9,9 +9,10 @@ const { container, legendText } = require("../../style/subpage.module.css");
 
 const { pamSite, mutation } = require("../../style/editing-design.module.css");
 
-const formatTextWithGeneLocations = (text: string, className: string) => {
+const formatTextWithGeneLocations = (text: string | {sequence: string} , className: string) => {
     // PAM sites and mutations are indicated in the string using square brackets
-    const parts = text.split(/(\[.*?\])/);
+    const input = typeof text === "string" ? text : text.sequence;
+    const parts = input.split(/(\[.*?\])/);
```
interim17, line 15: You can leave this function with a single string argument and no type check if you pass in `sequence.sequence` on ln60

interim17 (commented):
Screenshot 2025-03-10 at 10 21 16 AM

Do we have guidance from Travis on formatting these `Mutant`/`WT` sequence types on the strings? looks to me like they would want at least a colon if not some other formatting?

rugeli:
Hey @interim17 , here is how the table looks like in the [CofA pdf](https://www.coriell.org/0/PDF/Allen/iPSC/AICS-0107_CofA.pdf) for your reference. I can also double-check the sequence formatting with Travis.

Screenshot 2025-03-10 at 1 37 33 PM

src/components/SubPage/EditingDesign.tsx
```diff
@@ -43,10 +43,24 @@ const EditingDesignSubpage: React.FC<EditingDesignSubpageProps> = ({
     }
 
     if (dnaDonorSequence) {
+        const sequences = Array.isArray(dnaDonorSequence)
+            ? dnaDonorSequence
+            : [dnaDonorSequence];
```
meganrm, line 48: this shouldn't really be necessary right? you've now converted everything to an array

meganrm (approved):

src/components/SubPage/EditingDesign.tsx
rugeli: oh good catch! Will fix it now.

interim17 (approved):

rugeli:
Todo: ask scientists about the asterisk @rugeli

PR #123, opened 2025-03-26
responsive design at tablet scale

### LINKED ISSUE #114: Main image padding should go away at certain sizes
## Use Case
Currently the main image in the media gallery scales down a lot when on smaller screens 

## Acceptance Criteria

should look like the responsive design up to 744 (not the mobile design )

Padding and margin around image shrink first, but will always retain a minimum of 20px on all sides of image.
When 20px limit is reached, image shrinks while retaining aspect ratio.
 For full-screen designs, image width does not exceed 75% of the card containing it. Carousel is not considered part of image

## Details

Image

### LINKED ISSUE #115: the 75% width on the subpage content should be a max width
## Acceptance Criteria

on smaller screens the content should go full width 
below 1280 
any images that were side by side should stack 

## Details

### LINKED ISSUE #116: On tablets the image gallery should be below the main card
## Use Case

tablets are not wide enough to show data side by side 

## Acceptance Criteria
the content should be max width: 960px
the image gallery is below the main card with 20 px padding between them 
the main card should go full width 

## Details

interim17:
Problem
=======
Closes #115 Closes #114 Closes #116 

Solution
========
Seemed easiest to tackle a number of these issues at once since they rely on the same media queries and a lot of the rules/selectors interact with each other.

A few places that had `Flex` containers or inline styling i bumped into stylesheets as that makes the rules-in-question easier to target with media queries.

Haven't done a final review with Travis yet, still want to ask him about the 

### DIFF
--- a/src/style/disease-cell-line.module.css
+++ b/src/style/disease-cell-line.module.css
@@ -1,16 +1,45 @@
 .container {
-    max-height: 850px;
-    height: 100%;
     box-sizing: border-box;
+    display: grid;
+    grid-template-columns: auto 1fr;
+    width: 100%;
+    margin: 0 auto;
+    gap: 40px;
 }
 
 .left-card {
-    max-width: 505px;
-    width: 40%;
-    min-width: 460px;
-    flex: 1;
+    width: 460px;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    gap: 16px;
 }
+
 
 .return-arrow {
     margin-right: 8px;
+}
+
+/* smaller sizes on computer screen */
+@media screen and (min-width: 1280px) and (max-width: 1920px) {
+    .container {
+        padding: 0 20px;
+    }
+}
+
+/* tablet sizing */
+@media (max-width: 1279px) {
+    .container {
+        display: flex;
+        flex-direction: column;
+        max-width: 960px;
+        gap: 20px;
+        padding: 0 20px;
+        min-height: auto;
+    }
+
+    .left-card {
+        width: 100%;
+        min-height: auto;
+    }
 }
--- a/src/style/disease-cell-line.module.css
+++ b/src/style/disease-cell-line.module.css
@@ -20,13 +20,6 @@
     margin-right: 8px;
 }
 
-/* smaller sizes on computer screen */
-@media screen and (min-width: 1280px) and (max-width: 1920px) {
-    .container {
-        padding: 0 20px;
-    }
-}
-
 /* tablet sizing */
 @media (max-width: 1279px) {
     .container {
@@ -34,12 +27,16 @@
         flex-direction: column;
         max-width: 960px;
         gap: 20px;
-        padding: 0 20px;
         min-height: auto;
+        margin-bottom: 40px;
     }
 
     .left-card {
         width: 100%;
         min-height: auto;
+    }
+
+    :global(.ant-divider) {
+        display: none;
     }
 }
--- a/src/style/disease-cell-line.module.css
+++ b/src/style/disease-cell-line.module.css
@@ -5,6 +5,7 @@
     width: 100%;
     margin: 0 auto;
     gap: 40px;
+    margin-bottom: 40px;
 }
 
 .left-card {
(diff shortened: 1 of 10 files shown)


src/style/disease-cell-line.module.css
```diff
@@ -1,16 +1,42 @@
 .container {
-    max-height: 850px;
-    height: 100%;
     box-sizing: border-box;
+    display: grid;
+    grid-template-columns: auto 1fr;
+    width: 100%;
+    margin: 0 auto;
+    gap: 40px;
```
interim17, line 7: In general prefer flex, but since we can pivot back to flex when we drop below certain breakpoints, grid is actually really nice here to keep the containers the same height.

src/templates/disease-cell-line.tsx
```diff
@@ -49,8 +50,8 @@ export const DiseaseCellLineTemplate = ({
         (imagesAndVideos?.videos?.length || 0) > 0;
     return (
         <>
-            <Flex className={container} gap={40} justify="space-between">
-                <Flex vertical gap={16} className={leftCard}>
+            <div className={container}>
+                <div className={leftCard}>
```
interim17, line 54: antd `Flex` components are harder to use when altering style rules in media queries

src/style/disease-cell-line.module.css
```diff
@@ -1,16 +1,42 @@
...
 
+
 .return-arrow {
     margin-right: 8px;
-}
\ No newline at end of file
+}
+
+/* tablet sizing */
+@media (max-width: 1279px) {
```
meganrm, line 24: this should be 1280 or the other one should be 1279. if there is a discrepancy in the design you can ask Travis, but it's better to keep all the breakpoints consistent

src/templates/disease-cell-line.tsx
rugeli: Yeah I hear you. I often find myself struggle to tell when to use Ant Design versus CSS. I'll make a note to bring this up in the next group meeting for clarification.

src/style/disease-cell-line.module.css
interim17: Yeah these should have just been 1280, also just removed divider from all designs re: new feedback from Travis

src/style/images-and-videos.module.css
```diff
@@ -46,16 +47,19 @@
...
 .primary-image-with-thumbnail {
-    width: 80%;
+    max-width: 80%;
     flex: 1 1 auto;
 }
 
+.container .primary-image-container :global(.gatsby-image-wrapper) {
+    height: 0;
+}
+
```
rugeli, line 62: I tested with AICS-107 and 106 and found that when there's only one image, this rule makes the container collapse on full screen. Should we add a conditional class for single images or something?

interim17: hmm, thanks for catching that, the use of this rule was that it was letting the right panel height be constrained by the left panel without hard coding max values.

need a fix for that.

rugeli (approved):

src/style/images-and-videos.module.css
```diff
@@ -82,4 +86,60 @@
...
+    }
+
+    .primary-image-with-thumbnail {
+        max-width: 100%;
+        flex: 1 1 auto;
+    }
+}
+
+/* tablet sizing */
+@media (max-width: 1280px) {
```
meganrm, line 112: I don't understand why you have two

src/style/images-and-videos.module.css
```diff
@@ -82,4 +86,60 @@
...
+    }
+
+    .container :global(.ant-card-body) {
+        margin: 20px 0px;
+        gap: 20px;
+        justify-content: space-between;
+    }
+
+    .caption {
+        max-height: auto;
```
meganrm, line 130:
```suggestion
        max-height: min-content;
```
this is not working on chrome

src/style/images-and-videos.module.css
```diff
@@ -82,4 +86,60 @@
...
+    }
+
+    .primary-image-container {
+        gap: 20px;
```
meganrm, line 134: this isn't getting applied

meganrm (approved):

PR #124, opened 2025-03-31
Update data handling for gene, tag_location, fluorescent_tag arrays 

### LINKED ISSUE #97: Update queries and types to support multiple genes and tags
## Use Case

With the recent update allowing multiple entries for genes in PR #92, the queries and types need to be updated to correctly render multiple genes and tags.

Fields are being treated as strings but should be updated to arrays:
- `tag_location`
- `fluorescent_tag`
- `gene` 

## Acceptance Criteria
- Queries correctly retrieve and display multiple genes and tags 
- UI shouldn't be broken by the new display (if any)

rugeli:
Problem
=======
closes #97 

Solution
========
- update query config: return gene as an array instead of an object, tag_location and fluorescent_tag are also arrays
- for disease cell lines and their parental lines, display the first gene's data
- for normal cell lines, show all gene symbols, names, and locations 

## Type of change

Steps to Verify:
----------------
1. Check the preview, particularly multi-gene cell lines (AICS-59, 78, 84, 86, 89)

Screenshots (optional):
-----------------------
(TODO: Follow up with UX about multi-tag styling)
Screenshot 2025-03-31 at 4 40 18 PM

### DIFF
--- a/src/component-queries/convert-data.ts
+++ b/src/component-queries/convert-data.ts
@@ -7,12 +7,6 @@ import {
     UnpackedNormalCellLine,
 } from "./types";
 
-const multipleValues = (value: string | string[] | undefined): string => {
-    if (!value || value.length === 0) return "";
-    if (typeof value === "string") return value;
-    return value.length > 1 ? value.join(", ") : value[0];
-};
-
 
 const extractGenes = (geneArray: { frontmatter: GeneFrontMatter }[] = []): UnpackedGene[] => {
     if (!geneArray) return [];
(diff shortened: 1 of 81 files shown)


src/component-queries/convert-data.ts
```diff
@@ -2,13 +2,36 @@ import {
...
     UnpackedNormalCellLine,
 } from "./types";
 
+
+const extractGenes = (geneArray: { frontmatter: GeneFrontMatter }[] = []): UnpackedGene[] => {
+    if (!geneArray) return [];
+
+    return geneArray
+        // TODO: check with scientists
+        // filter out null values, e.g. UBTF, CLYBL-dCas9-KRAB
```
meganrm, line 16: this should be removed, you can make it a todo, but you shouldn't be able to add a gene to a cell line without first adding it to the collection

src/components/CellLineTable/NormalTableColumns.tsx
```diff
@@ -13,6 +13,7 @@ import {
     certificateOfAnalysisColumn,
     obtainLineColumn,
 } from "./SharedColumns";
+import { render } from "react-dom";
```
meganrm, line 16: I think this is a mistake?

rugeli: Thanks for catching it!

rugeli:
Follow up ticket: Unify fields under a parent gene editing list #134

meganrm (approved):

Issue #2, opened 2024-02-22
get the parental line to successfully unpack onto the disease cell line


meganrm:
# Bug report

**What is the current behavior?**
The project won't build because the parental line isn't formatting correctly 

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
The data needed for the parental line modal should be on the disease cell line data

**Other relevant information:**

Node.js version: 
NPM/Yarn version
Operating System: 
Additional tools:

PR #162, opened 2025-05-12
Feature/table sort

### LINKED ISSUE #141: Sort on columns
## Use Case

Filter has been stripped down and search and sort have been added. This ticket describes how sort works for the main table.

## Details

Columns are sortable- [see prototype for interaction](https://www.figma.com/proto/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?page-id=235%3A2245&node-id=4926-7185&viewport=-5294%2C629%2C0.35&t=lDprALX8g0jJmkQf-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=4926%3A7185). By default they are sorted on AICS ID. Only one column can be sorted at a time and a sort cannot be removed, only changed to sorting a different column (e.g. back to AICS ID).

Each column label has a caret indicating direction of sort. Columns that are sorted on get a semibold label and caret changes to same color as label (#323233). [Explore the raw component and its prototyping here.](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=970-1880&t=GYUrxyq4ZXHiW2rJ-1)

interim17:
Problem
=======
Closes #141 and #165

Solution
========

Add sorter functions to each column object in `getNormalTableColumns`
Implement state, change handler, and necessary details for styling in `CellLineTable`

Screenshot 2025-05-12 at 10 20 50 AM

### DIFF
--- a/src/components/CellLineTable/NormalTableColumns.tsx
+++ b/src/components/CellLineTable/NormalTableColumns.tsx
@@ -137,11 +137,5 @@ export const getNormalTableColumns = (
             { ...certificateOfAnalysisColumn },
         ];
     }
-    // remove the sorting function from the inprogress table
-    return columns.map((column) => {
-        return {
-            ...column,
-            sorter: undefined,
-        };
-    });
+    return columns;
 };
--- a/src/components/CellLineTable/index.tsx
+++ b/src/components/CellLineTable/index.tsx
@@ -1,7 +1,5 @@
 import React, { useState } from "react";
 import { Table, Tag, Flex } from "antd";
-import { CaretDownOutlined, CaretUpOutlined } from "@ant-design/icons";
-import { SorterResult, SortOrder } from "antd/es/table/interface";
 import { navigate } from "gatsby";
 
 import { HTMLContent } from "../shared/Content";
@@ -19,8 +17,6 @@ const {
     footer,
     hoveredRow,
     dataComplete,
-    columnHeader,
-    activeColumn,
 } = require("../../style/table.module.css");
 
 interface CellLineTableProps {
@@ -89,6 +85,7 @@ const CellLineTable = ({
         }
         return {
             ...column,
+            sorter: inProgress ? undefined : column.sorter,
             onCell: inProgress ? undefined : onCellInteraction,
         };
     });


src/components/CellLineTable/NormalTableColumns.tsx
```diff
@@ -16,17 +16,26 @@ import {
 
 const { lastColumn } = require("../../style/table.module.css");
 
+/* case insensitive string comparison */
+const compareStrings = (a = "", b = "") =>
```
meganrm, line 20: nit, but could call this `caseInsenstiveStringCompare` and then you don't need the comment

meganrm:
when I click the arrow a third time it does something weird

meganrm:
should the down arrow for cell id be highlighted on load since that's what's being sorted on by default? (the prototype travis made isn't loading for me right now so I can't see the design)

src/components/CellLineTable/index.tsx
```diff
@@ -37,6 +41,22 @@ const CellLineTable = ({
     mobileConfig,
 }: CellLineTableProps) => {
     const [hoveredRowIndex, setHoveredRowIndex] = useState(-1);
+    const [sortedColumn, setSortedColumn] = useState<{
+        key: string;
+        order: SortOrder;
+    }>({ key: "", order: null });
```
meganrm, line 47: isn't it sorted on id by default?

src/components/CellLineTable/index.tsx
```diff
@@ -37,6 +41,22 @@ const CellLineTable = ({
...
+
+    const onSortingChange = (
+        _p: any,
+        _f: any,
+        sorter: SorterResult<UnpackedCellLine> | SorterResult<UnpackedCellLine>[]
+    ) => {
+        const s = Array.isArray(sorter) ? sorter[0] : sorter; // multi‑sort guard
+        const sortedKey = s.columnKey?.toString() ?? "";
+        const sortedOrder = s.order ?? null;
+        setSortedColumn({ key: sortedKey, order: sortedOrder });
```
meganrm, line 57: it's just a guess, but maybe the weird behavior I'm seeing is because you're saving an object here. maybe just have two separate pieces of state: 
`setSortedColumnKey`
`setColumnSortOrder`

interim17: Yes the arrow should be active on the ID column right away, I had that wrong. 

Changed initial state values to match.

Thanks for catching bug, the third click behavior is actually because the antd sort directions array has a third undefined option after ascend and descend. The pushed changed on ln118 should fix it.

If you want me to change the state to two separate variables I am happy to but I don't think that was the source of the bug.

meganrm: no it's not necessary

src/components/CellLineTable/index.tsx
```diff
@@ -77,12 +99,33 @@ const CellLineTable = ({
...
         // the two clickable columns are the order cell line and
         // CoA column. They do not have the hover effect and
-        // should not take you to the cell line page
+        // should not take you to the cell line page, and are not
+        // sortable.
         if (column.className?.includes("action-column")) {
             return column;
         }
+        const isSortedColumn = sortedColumn.key === column.key;
+        const sortIcon =
```
rugeli, line 108: non-blocking: from my personal user experience, adding tooltips like `Sort A-Z` or `Ascending`/`Descending` to the sort icon would be very helpful.

rugeli:
Also noticing the clicking issues Megan mentioned:
1. For the `fluorescent tag`, `tag location` columns, there seems to be a "no reaction" on every third click (pattern: up, down, nothing happens)
2. For `protein`, `gene symbol & name` and `structure` columns, there's a similar issue but slightly different - on that third click, the sorting actually happens but the arrow direction doesn't update to match
2.5 nit: the sorting pattern seems a bit off, for example, clicking three times on protein, the top one will go `paxillin`-`alpha-actinin-2`- `Vascular endothelial (VE)-cadherin`. This may resolve itself once we fix #2.

interim17:
> Also noticing the clicking issues Megan mentioned:
> 
> 1. For the `fluorescent tag`, `tag location` columns, there seems to be a "no reaction" on every third click (pattern: up, down, nothing happens)
> 2. For `protein`, `gene symbol & name` and `structure` columns, there's a similar issue but slightly different: on that third click, the sorting actually happens but the arrow direction doesn't update to match
>    2.5 nit: the sorting pattern seems a bit off, for example, clicking three times on protein, the top one will go `paxillin`-`alpha-actinin-2`- `Vascular endothelial (VE)-cadherin`. This may resolve itself once we fix 2.
> 
> (Just saw the latest commit for fixing the undefined third click, looks like it still persists though? Happy to help test more if needed!)

I only added the value for the third click on the ID column initially, I just added it to the other array so that all columns have all three values, looks like that might fix it?

rugeli (approved):
The fix looks good! Bugs I've seen are all cleared up!

src/components/CellLineTable/index.tsx
```diff
@@ -74,17 +98,51 @@ const CellLineTable = ({
...
+        const interactiveColumn = {
             ...column,
             onCell: onCellInteraction,
         };
+        if (sortableTable && column.sorter) {
+            return applySortingProperties(interactiveColumn);
+        }
+        return {
+            ...interactiveColumn,
+            sorter: undefined,
```
meganrm, line 144: why do you have to do this? shouldn't it already be undefined if it's not a sortable column?

src/components/CellLineTable/index.tsx
```diff
@@ -74,17 +98,51 @@ const CellLineTable = ({
...
         if (column.className?.includes("action-column")) {
             return column;
         }
-        return {
+        const interactiveColumn = {
             ...column,
             onCell: onCellInteraction,
         };
+        if (sortableTable && column.sorter) {
+            return applySortingProperties(interactiveColumn);
```
meganrm, line 140: can't you just check if it has a `sorter`?

src/components/CellLineTable/NormalTableColumns.tsx
```diff
@@ -16,17 +16,25 @@ import {
...
+
 export const getNormalTableColumns = (
     inProgress: boolean
 ): CellLineColumns<UnpackedNormalCellLine> => {
     const columns = [
-        { ...cellLineIdColumn },
+        {
+            ...cellLineIdColumn,
+            sorter: (a: any, b: any) =>
+                (a.cellLineId ?? 0) - (b.cellLineId ?? 0),
```
meganrm, line 29: I would move the sort order setting here. the other things you're doing to the columns are all based on on state changes, but this is an initial setting so I would put it in this document. 
two suggestions:
1. `const defaultSortOrder = ["ascend", "descend", "ascend"]` and `const reverseSortOrder = ["v", "ascend", "descend"] and add the `sortDirections` attribute here. You could put the actual definitions in a shared file so if/when we add sorting to the Disease table they're accessible. 
2. add the default icon that goes next to the title here (same reason, it's an initial setting). If you don't want to write the same jsx a lot, you could map over the columns before you export them to check for a sorter and add the icon. 
Also: since the antd icon is already a triangle, couldn't you use that, and then just use css to rotate it? instead of using two separate icons and having to check to see if it's a sorter. it just seems like it would eliminate a lot of conditional code here

src/components/CellLineTable/index.tsx
```diff
@@ -110,6 +168,8 @@ const CellLineTable = ({
                 expandable={isTablet ? mobileConfig : undefined}
                 columns={interactiveColumns}
                 dataSource={cellLines}
+                onChange={sortableTable ? onSortingChange : undefined}
```
meganrm, line 171: I don't think you need this check

src/components/CellLineTable/NormalTableColumns.tsx
meganrm: actually I don't think you need these two separate lists if you just use `defaultSortOrder` on the cell line id column

interim17: So I just realized we can actually set the default sort directions on the table itself as a prop, then I can add the reverse here only for the ID column, and I don't need to duplicate it on the other columns or have the conditional check in the component so thats nice.

meganrm (approved):

PR #169, opened 2025-05-12
cell catalog footer update

### LINKED ISSUE #164: Acknowledgements and Funding footer
## Use Case

Update text and design in footer to match [design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=4975-10813&t=SHn1ecxcZ0WZ9mmA-4).

interim17:
Problem
=======
Closes #164

Solution
========

with @pairperson1

## Type of change

* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* This change requires a documentation update
* This change requires updated or new tests

Steps to Verify:
----------------
1. A setup step / beginning state
2. What to do next
3. Any other instructions
4. Expected behavior
5. Suggestions for testing

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

Keyfiles (delete if not relevant):
-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/style/footer.module.css
+++ b/src/style/footer.module.css
@@ -1,0 +1,20 @@
+.footer {
+    color: #204A6C;
+    display: flex;
+    flex-direction: column;
+    font-size: 16px;
+    font-weight: 400;
+    gap: 24px;
+}
+
+.footer p {
+    margin: 8px 0px;
+}
+
+.footer .contributorsClass .contributor {
+    font-weight: 600;
+}
+
+.footer .contributorsClass .institution {
+    font-style: italic;
+}
--- a/src/templates/normal-catalog.tsx
+++ b/src/templates/normal-catalog.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { Card, Divider, Flex } from "antd";
 import { graphql } from "gatsby";
 import Layout from "../components/Layout";
+import Footer from "../components/CellLineTable/Footer";
 import Content, { HTMLContent } from "../components/shared/Content";
 import { GatsbyImage, getImage } from "gatsby-plugin-image";
 import { FileNode } from "gatsby-plugin-image/dist/src/components/hooks";
@@ -19,7 +20,12 @@ interface NormalCatalogTemplateProps {
     title: string;
     content: string;
     contentComponent?: JSX.ElementType;
-    footerText: string;
+    fundingText: string;
+    acknowledgementsBlock: {
+        intro: string;
+        contributors: { name: string; institution: string }[];
+        outro: string;
+    };
     main: {
         heading: string;
         description: string;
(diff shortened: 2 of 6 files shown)


src/style/footer.module.css
```diff
@@ -0,0 +1,20 @@
+.footer {
```
rugeli, line 1: nit: the line spacing in `funding` section looks larger than in intro/outro of `acknowledgments` by eye. Could this be `HTMLContent` renders funding text differently?

src/templates/normal-catalog.tsx
```diff
@@ -87,6 +97,14 @@ interface QueryResult {
                 footer_text: {
                     html: string;
                 };
```
rugeli, line 99: remove?

src/style/footer.module.css
```diff
@@ -0,0 +1,20 @@
...
+    color: #204A6C;
```
meganrm, line 2: is this a new color not in our color vars?

src/style/table.module.css
```diff
@@ -117,10 +117,10 @@
 }
 
 .footer {
-    padding: 8px 0 16px 0;
     font-size: 14px;
     font-weight: 300;
     color: var(--ALLEN_LIGHT_60);
+    margin-top: 45px;
```
meganrm, line 123: why this change? are you using this class anywhere?

src/style/footer.module.css
```diff
@@ -0,0 +1,20 @@
...
+    font-size: 16px;
+    font-weight: 400;
+    gap: 24px;
+}
+
+.footer p {
+    margin: 8px 0px;
+}
+
+.footer .contributorsClass .contributor {
```
meganrm, line 14: there's no real reason for this contributorsClass as far as I can see, did you need it for selector specificity? (also, nit, but having `class` in a class name is a little redundant)

src/style/footer.module.css
interim17: it's not, i'll add it. I'm thinking.. "maritime blue"

src/style/table.module.css
interim17: It's used on ln 114 in CellLineTable.

Goal is to get spacing correct between table and footer.

It seemed to me that that bottom padding rule above wasn't being applied meaningfully. The top padding was being applied but was too small, and It seems more appropriate to use margin for this, so I removed the padding rule for clarity and used margin instead.

src/style/footer.module.css
meganrm: the colors also have names in figma, like blue_80, I've been using those so we will be able to recognize it when they're used again

interim17: Oh that's amazing! I've always wanted a figma to repo naming match on color vars

meganrm (approved):

PR #170, opened 2025-05-13
Fix: cell catalog footer data

rugeli:
Problem
=======
secondary brach for #169, merging into feature/footer

Solution
========
- fixed graphql query error in disease catalog
- adjusted footer data format for cell catalog
- made cell catalog `subheading` optional to prevent data updates submission error when this field is empty

## Type of change

Steps to Verify:
----------------
1. All pages should load properly

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -11,6 +11,7 @@ exports.createSchemaCustomization = ({ actions, schema }) => {
             gene: [MarkdownRemark] @link(by: "frontmatter.symbol", from: "gene") 
             parental_line: MarkdownRemark @link(by: "frontmatter.cell_line_id")
             funding_text:  String @md
+            footer_text: String @md
             acknowledgements: String @md
         }`,
     ];
--- a/static/admin/config.yml
+++ b/static/admin/config.yml
@@ -738,23 +738,23 @@ collections:
                 - { label: "Title", name: "title", widget: "string" }
                 - { label: "Body", name: "body", widget: "markdown" }
                 - {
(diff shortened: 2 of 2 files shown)


gatsby-node.js
```diff
@@ -11,6 +11,7 @@ exports.createSchemaCustomization = ({ actions, schema }) => {
             gene: [MarkdownRemark] @link(by: "frontmatter.symbol", from: "gene") 
             parental_line: MarkdownRemark @link(by: "frontmatter.cell_line_id")
             funding_text:  String @md
+            footer_text: String @md
```
interim17, line 14: In my branch I was aiming to replace/remove `footer_text`, is it still used somewhere such that we need to keep this in?

interim17 (approved):
Left question about restoring `footer_text` in the schema

gatsby-node.js
rugeli: this is to support existing queries in disease catalog component, preventing the error below:
```
There was an error in your GraphQL query:

      Field "footer_text" must not have a selection since type "String"
has no subfields.

      This can happen if you e.g. accidentally added { } to the field
"footer_text". If you didn't expect "footer_text" to be of type "String"
 make sure that your input source and/or plugin is correct.
      However, if you expect "footer_text" to exist, the field might be
accessible in another subfield. Please try your query in GraphiQL.

      It is recommended to explicitly type your GraphQL schema if you
want to use optional fields.

File: src/templates/disease-catalog.tsx:127:29
```

PR #175, opened 2025-05-19
Feature/search

### LINKED ISSUE #166: Add search to main cell catalog table
## Use Case

Implement search functionality and search bar UI that matches [design](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=5022-20066&t=SHn1ecxcZ0WZ9mmA-4).

## Acceptance Criteria

Search bar returns cell lines with any partial value included in table. Hitting clear resets search.

meganrm:
Problem
=======
closes #166 

Solution
========
The search is an autocomplete to give users feedback on what is available as they type. 

## Type of change

Steps to Verify:
----------------
1. go to the catalog
2. search for something 
3. you can select a specific item from the dropdown, or you can hit enter and it will search the value you've typed


meganrm:
just found a bug, stay tuned

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+                }}
+            >
+                <Input placeholder="Search AICS ID, protein, gene..." />
+            </AutoComplete>
+            <GrayButton
+                onClick={handleClear}
+                className={currentValue ? active : ""}
+            >
+                Clear
+            </GrayButton>
```
ShrimpCryptid, line 134: Should this be disabled?

src/style/buttons.module.css
```diff
@@ -34,3 +34,17 @@
...
+    background-color: var(--GRAY_30) !important;
+    border: none !important;
+    color: var(--WHITE) !important;
+}
+
+.gray-button.active,
+.gray-button.active:hover {
+    background-color: var(--primary-color) !important;
+    border: none !important;
+    color: var(--WHITE) !important;
```
ShrimpCryptid, line 49: May not be necessary since L41-42 will still apply

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+        setCurrentValue("");
+    };
+
+    const onEnter = () => {
+        let selectedCellLines = new Set<number>();
+        options.forEach((option) => {
+            selectedCellLines = new Set([
+                ...selectedCellLines,
+                ...getCellLineIdsToShow(option.value),
+            ]);
```
ShrimpCryptid, line 52: Can you add the elements to the initialized Set instead of creating a new Set every time?

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+                onSelect={onSelect}
+                onSearch={handleSearch}
+                allowClear
+                onClear={() => {
+                    setOptions([]);
+                    setResults(allCellLines);
+                }}
+                onInputKeyDown={(e) => {
+                    if (e.key === "Enter") {
+                        e.preventDefault();
```
ShrimpCryptid, line 122: What does the default do here?

src/components/SearchBar.tsx
meganrm: I could, this was a bit of a remnant from how I was doing things previously where there were a lot of duplicates that needed to be removed and this is a quick way to de-dup a list. I don't feel strongly about either method (this way verses iterating over the returned array and adding each element)

interim17 (commented):
Feature is working and looks good in UI, left some questions about the overall approach of using a new graphql query for this feature.

My gut says using the unpacked data would be better than a new query but I'd like to hear your thoughts about it.

src/component-queries/SearchAndFilter.tsx
```diff
@@ -0,0 +1,85 @@
...
+    setResults,
+}: SearchAndFilterProps) => {
+    return (
+        <StaticQuery
+            query={graphql`
+                query SearchAndFilterQuery {
+                    allMarkdownRemark(
+                        filter: {
+                            frontmatter: {
+                                templateKey: { ne: "disease-cell-line" }
```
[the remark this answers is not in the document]
meganrm: the reason is to reduce the number of times we're mapping over the data. all the queries are generated server side, so the more we can do with queries the faster the front end is.

src/component-queries/SearchAndFilter.tsx
```diff
@@ -0,0 +1,85 @@
...
+                                cell_line_id: { ne: 0 }
+                                status: { ne: "hide" }
+                            }
+                        }
+                    ) {
+                        group(
+                            field: {
+                                frontmatter: {
+                                    gene: { frontmatter: { symbol: SELECT } }
```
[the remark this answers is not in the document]
meganrm: it maps gene ->cell lines. I might have to adjust it a little bit once Ruge's changes are in, but it's every gene element. the issue is here, the most likely use case is users searching for genes or proteins, so I didn't want to have to go through the array of cell lines on every search to see if a gene matches, instead I have a lookup map of genes that returns cell lines.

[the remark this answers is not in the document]
interim17: Makes sense! I have another (non-blocking) strategy question.

It's nice to have the query live with its component because it makes everything easy to understand when looking at this file.

But if there are schema changes that break queries, typescript isn't really helping us because these graphql queries are just strings, so we have to manually go track down everything that might break. Is there a better way to do this so that schema<->query relationships could be more maintainable? Testing would be one way I guess.

interim17 (approved):
Left one more non blocking strategy question about where queries should live but this LGTM

src/components/SearchBar.tsx
ShrimpCryptid: This works fine on Windows 11 Chrome!

ShrimpCryptid (approved):

Issue #140, opened 2025-04-22
Main table content display specifications


TravisKroeker:
## Use Case

The fields for the main table have been slightly altered from the current cell catalog and layout requirements have been instituted in order to save space and make display more compact. 

## Details

Font is 14px, regular. I think disease catalog is 16px so there is a difference. Will think about reducing the disease cell catalog to match but leave it for now.

Columns for clone and parental line have been eliminated to save space. Clone is useless to a user when there is only one and the wtc parental line is the same for every entry. A note has been elsewhere about the parental line, it doesn't need to be displayed in the table.

Most columns are set so their width is identical to that of the largest text object in that column + 16px padding on each side (the largest text object could be the label). Some columns have a max width applied, after which the text is truncated. These columns are Protein (max width 220px, Gene symbol/name (max width 310px, which also includes the tag), and structure (max width 220px).

For any fields that are truncated, a tooltip appears on hover displaying the full name.

[See designs for Figma details and feel free to ask questions.](https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=4926-7185&t=GYUrxyq4ZXHiW2rJ-1)

PR #176, opened 2025-05-19
Feature/fontsize

### LINKED ISSUE #172: change font to 14px

interim17:
Problem
=======
Advances #140 
Closes #172 

Solution
========
Pulled normal cell line table name into constant since I'm using it to check which table is rendering and add a class name. 

Needed since disease table is staying at 16px and normal cell table is going to 14px;


src/components/CellLineTable/index.tsx
```diff
@@ -90,11 +92,17 @@ const CellLineTable = ({
         };
     });
 
+    const tableClassNames = classNames(
+        container,
+        inProgress ? comingSoon : "",
+        tableName === NORMAL_CELLS_TABLE_NAME ? normalCells : ""
+    );
```
meganrm, line 99: the idea of this component is that it's not supposed to really know/care what kind of table it is. I think a cleaner way to implement this is to put a class name in the `normal-catalog.tsx` template that targets font size or to use the `classNames` on columns

interim17: yeah that's much simpler!

meganrm (approved):

rugeli (approved):
LGTM!

PR #209, opened 2025-05-30
Render thumbnails for cell line main table

### LINKED ISSUE #167: include images in table
Have the static image display as it does in the design with no hover effect

rugeli:
Problem
=======
closes #167 

Solution
========
- added static thumbnail images to cell line main table

Next steps: 
- add hover effects
- move the rest of thumbnails over from the legacy cell catalog 

with @interim17 

Note: I removed `width: 164` from the first columns in this pr as there are other unmatched sizing in the table. Might be cleaner to take care of sizing issues in a separate pr.
Screenshot 2025-05-29 at 7 05 24 PM

## Type of change

### DIFF
--- a/src/component-queries/convert-data.ts
+++ b/src/component-queries/convert-data.ts
@@ -105,6 +105,7 @@ export const convertFrontmatterToNormalCellLines = ({
         healthCertificate: "",
         orderLink: cellLineNode.frontmatter.order_link,
         orderPlasmid: cellLineNode.frontmatter.donor_plasmid,
+        thumbnailImage: cellLineNode.frontmatter.thumbnail_image,
     };
 };
 
--- a/src/components/CellLineTable/DiseaseTableColumns.tsx
+++ b/src/components/CellLineTable/DiseaseTableColumns.tsx
@@ -1,12 +1,14 @@
 import React from "react";
 import { Flex } from "antd";
 import classNames from "classnames";
+import { Link } from "gatsby";
 
 import {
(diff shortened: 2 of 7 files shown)


src/components/CellLineTable/DiseaseTableColumns.tsx
```diff
@@ -1,31 +1,53 @@
...
+import { smBreakPoint, mdBreakpoint, CellLineColumns, UnpackedCellLine } from "./types";
 
 const {
     clones,
     lastColumn,
     snpColumn,
+    cellLineId,
 } = require("../../style/table.module.css");
 
+const cellLineIdColumn = {
```
meganrm, line 31: nit, but there is not need to define this separately and then make a copy of it below. you can just define it right in the array

src/component-queries/convert-data.ts
```diff
@@ -105,6 +105,7 @@ export const convertFrontmatterToNormalCellLines = ({
         healthCertificate: "",
         orderLink: cellLineNode.frontmatter.order_link,
         orderPlasmid: cellLineNode.frontmatter.donor_plasmid,
+        thumbnailImage: cellLineNode.frontmatter.thumbnail_image,
```
meganrm, line 108: I would unpack this all the way, so thumbnailImage type is `gatsbyImageData | null`

meganrm (approved):

interim17 (approved):

PR #218, opened 2025-06-04
remove unused main from normal catalog template and query

### LINKED ISSUE #182: remove `main` content from the normal catalog template and the query
## Use Case

the new design doesn't have the content for `main` 

## Acceptance Criteria

remove it from the template in the config file and from the query

interim17:
Problem
=======
Closes #182

Solution
========
Ticket is to remove the main section from query and component, but the text in main is actually still in use.
Follow up PR will point at this branch and replace the text "Allen Cell Collection Cell Lines" in new field in the markdown.


rugeli (approved):
nit: same here, we can clean up the data fields in the admin yml file to keep things clearer for future users. Otherwise lgtm!

meganrm:
looks fine but needs to resolve conflicts

PR #225, opened 2025-06-04
run proxy and app in one terminal command

interim17:
Problem
=======
Quality of life.

Solution
========
A script to run the proxy server and the app in a single command.

Added `concurrently` as a dev dependency, the script `npm run dev` does everything in one window.

Might like to add a watch flag with nodemon to rebuild on config changes in the future. npx and start scripts are unchanged if you don't want to use this but I am annoyed managing the two terminal windows


rugeli (approved):
I'm up to this idea! Since the proxy server is mostly quiet, the terminal outputs stay clean and easy to follow and I had no issues logging in admin page locally. Thanks for the change!

meganrm (approved):

PR #252, opened 2025-07-10
Fix/parent line bug

### LINKED ISSUE #198: Parental line popover bug
After opening a parental line popover, clicking anywhere on the popover (other than the open parental line button) causes the secondary page for the child line to open. Nothing should happen in this instance.

interim17:
Problem
=======
Closes #198 

Solution
========
Row click control and modal state don't share a direct parent, so best fix I found was to pass around a ref that the modal can set/unset and the table can listen to.

Alternate approaches I tried seemed unwieldy or to break our patterns on where data should live/be queried.


rugeli (approved):
I've tested it and it worked well!

meganrm (approved):

Issue #258, opened 2025-07-11
Cell Line subpage: Info panel


rugeli:
## Use Case

_Please provide a use case to help us understand your request in context_
Similar to disease line subpages, we need a cell line info card for normal lines.

## Acceptance Criteria

_Please describe how you know this is done_
- reuse `Link`, `CellLineInfoCard` and `DefaultButton` components
- in `CellLineInfoCard` remove parental line and the clone table, add a button for Obtain donor plasmid.

Issue #254, opened 2025-07-10
build out normal cell line pages


interim17:
## Acceptance Criteria

_Please describe how you know this is done_

## Details

Copy design from disease catalog and add/remove relevant fields

rugeli:
Subpage development checklist:

**CellLineInfoCard**

- [ ]   Migrate data from catalog 1.0
- [ ]   Add data fields to admin page
- [ ]   Implement info card

**ImagesAndVideos**

- [ ]   Migrate data from catalog 1.0
- [ ]   Add data fields to admin page
- [ ]   Display images and thumbnails
- [ ]   Validate video links
- [ ]   Display videos and thumbnails 
- [ ]   Implement media carousel 

**Tabs** 

- [ ]   Confirm tab list for 2.0 with ux 
- [ ]   Migrate data from catalog 1.0
  - [ ]     Editing design: str(table data) and img with caption
  - [ ]     Genomic characterization: img with caption and str(table data with caption)
  - [ ]     Cellular validation: img
  - [ ]     Stem cell characteristics: str(table data with caption) and img with caption
  - [ ]     Protocols: files(pdf)
  - [ ]     Supplementary information: files(pdf)
  - [ ]     STR analysis: files(pdf). Tab displays extracted data from the pdf
- [ ]   Add data fields to admin page
- [ ]   Implement each tab (separate pr per tab)

meganrm:
I'm going to close this issue. the protocols and sup info are non cell line specific so they're no longer going to be associated with a subpage

PR #264, opened 2025-07-14
normal cell line info card

interim17:
Problem
=======
Advances #254 Advances #258

Solution
========
Not considering this feature done-done until we get input from UX, but this is a decent WIP for the normal cell line info card.

Besides rendering the card I did a small bit of refactoring on the way we get the parental line data into the disease line card to increase type specificity and avoid some data duplication.

`CellLineInfoCardBase` is the primary component which defines base prop interface and a union type for Normal and Disease props.

`NormalCellLineIfnoCard` and `DiseaseCellLineInfoCard` are basically data layers to keep things DRY, if the two card designs diverge too much in the future we can scrap the base component.


src/component-queries/types.ts
```diff
@@ -219,7 +221,16 @@ export interface UnpackedNormalCellLine extends UnpackedCellLineMainInfo {
...
+export type ParentLine = Pick<UnpackedNormalCellLine,
+    "thumbnailImage" |
+    "cellLineId" |
+    "cloneNumber" |
+    "taggedGene" |
+    "alleleCount" |
+    "tagLocation" |
+    "fluorescentTag" |
+    "geneticModifications"
+>
```
interim17, line 233: We were using the raw snake case front matter instead of unpacked data in parts of `disease-cell-line.tsx`

The `Partial` utility type makes all properties optional, so to use those unpacked properties without optional chaining I made this type more specific.

src/components/SubPage/convert-data.ts
```diff
@@ -118,21 +121,29 @@ export const getStemCellCharData = (
...
+    return {
+        thumbnailImage,
+        cellLineId,
+        cloneNumber,
+        taggedGene,
+        alleleCount,
+        tagLocation,
+        fluorescentTag
+    }
+}
```
interim17, line 142: We were duplicating some data below (`parentalLine` as well as `parentalLineGene`), so factored this out for clarity.

src/templates/cell-line.tsx
```diff
@@ -70,16 +86,16 @@ export const CellLineTemplate = ({
 
 const CellLine = ({ data }: QueryResult) => {
     const { markdownRemark: cellLine } = data;
+
+    const unpackedCellLine = convertFrontmatterToNormalCellLines({
+        node: cellLine,
+    });
+
```
interim17, line 93: The disease cell line subpage uses a different unpacking function to deal with the subpage data, but we don't need that yet, this covers everything.

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -1,81 +1,78 @@
...
+    cloneNumber: number;
+}
+
+export interface InfoTableRow {
+    key: React.Key;
+    label: string;
+    children: ReactNode;
+}
+
+export type CellLineInfoCardProps = DiseaseProps | NormalProps;
```
rugeli, line 48: unused type, remove?

rugeli (approved):
Thanks for cleaning up the types and adding shared components, the info card looks great! I merged one commit to add more data to the card, but like you said, it’s not finalized until we see the design. I left two comments for small adjustments, otherwise lgtm!

meganrm:
the build is failing, were you able to build this locally (not just the dev server, but actually run build?)

src/templates/cell-line.tsx
```diff
@@ -1,64 +1,88 @@
...
+                <Link to="/normal-catalog">
+                    <DefaultButton>
+                        <Arrow className={returnArrow} />
+                        Return to Cell Catalog
+                    </DefaultButton>
+                </Link>
+                <NormalCellLineInfoCard
+                    href={href}
+                    cellLineId={cellLineId}
+                    geneName={taggedGene[0].name}
```
meganrm, line 70: why only the first one? this should be the same as the table, where if they have more than one gene they are listed out

src/components/CellLineInfoCard/NormalCellLineInfoCard.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import CellLineInfoCardBase, { NormalProps } from "./CellLineInfoCardBase";
+
+export const NormalCellLineInfoCard: React.FC<NormalProps> = (props) => {
+    console.log("NormalCellLineInfoCard props:", props);
```
meganrm, line 5: remove log

meganrm:
https://deploy-preview-264--cell-catalog.netlify.app/cell-line/AICS-5-50/

this is now building, but I'm not sure why the bot hasn't updated the link

src/templates/cell-line.tsx
meganrm: I'm not saying render a table, I'm saying render it like we do on the table, 
Screenshot 2025-07-22 at 11 34 16 AM

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -1,81 +1,80 @@
...
-    parentalLine: ParentalLineFrontmatter;
-    parentLineGene: GeneFrontMatter;
     clones: Clone[];
+}
+
+export interface NormalProps extends BaseCellLineProps {
+    orderPlasmid: string;
+    cloneNumber: number;
+    taggedGene: { name: string; symbol: string; structure?: string; protein?: string }[];
+    protein: string[];
```
meganrm, line 41: why is protein in here twice? (taggedGene)

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -1,81 +1,80 @@
...
+}
+
+export interface CardLayoutProps {
+    href: string;
+    cellLineId: number;
+    infoRows: InfoTableRow[];
+    certificateOfAnalysis: string;
     healthCertificate: string;
+    orderLink: string;
+    cloneSummary?: { numMutants: number; numIsogenics: number };
```
meganrm, line 59: this should really be part of the disease card only, not the generic props

src/style/cell-line.module.css
```diff
@@ -1,11 +1,37 @@
...
+    display: grid;
+    grid-template-columns: auto 1fr;
+    width: 100%;
+    margin: 0 auto;
+    gap: 40px;
+    margin-bottom: 40px;
+}
+
+.left-card {
+    width: 460px;
```
meganrm, line 12: are these really necessary? can't we reuse the style sheet from the disease subpage?

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
interim17: The approach I took was to have a base component, and then two data layers.

If we don't include everything that might get rendered in this layout props, then we can't really have a base component, we will need two completely separate card components.

I'll refactor if you prefer that approach, was aiming to not duplicate the shared JSX/styling if possible.

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
interim17: Ah yeah, I can remove it from this taggedGene object in the props, but the data is still duplicated in the `convertFrontmatterToNormalCellLines` function (unchanged so far in this PR).

src/templates/cell-line.tsx
meganrm: I think the TS code should be written in a way that we won't have to change it, ie, we know we're going to render all of them. then in the future you should just be able to adjust the css

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
meganrm: I would think you'd remove it from the outer props, and leave it in tagged genes, since they're all related

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -1,81 +1,80 @@
...
+
+export interface DiseaseProps extends BaseCellLineProps {
+    parentalLine: ParentLine;
     snp: string;
-    orderLink: string;
-    certificateOfAnalysis: string;
-    parentalLine: ParentalLineFrontmatter;
-    parentLineGene: GeneFrontMatter;
     clones: Clone[];
+}
```
meganrm, line 35: our convention is to have the props right above and in the same file as the component that is using those props

src/components/CellLineInfoCard/DiseaseCellLineInfoCard.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import CellLineInfoCardBase, { DiseaseProps } from "./CellLineInfoCardBase";
+import { getCloneSummary, formatCellLineId } from "../../utils";
+
+export const DiseaseCellLineInfoCard: React.FC<DiseaseProps> = (props) => {
```
meganrm, line 5: similarly : rename to DiseaseCellLineInfoCardProps

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -127,6 +125,15 @@ const CellLineInfoCard = ({
...
                     hPSCreg Certificate
                 </DefaultButton>
+                {orderPlasmid && (
+                    <DefaultButton
+                        href={orderPlasmid}
+                        target="_blank"
+                        rel="noreferrer"
+                    >
+                        Obtain Plasmid
+                    </DefaultButton>
```
meganrm, line 135: if I was trying to make this a generic component, I would make a `buttonList: {label: string, href: href}` and then this component doesn't need to know about props like `orderPlasmid`

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -110,8 +108,8 @@ const CellLineInfoCard = ({
     );
     return (
         <Card title={titleContents} className={container}>
-            <InfoPanel data={tableData} />
-            <CloneTable dataSource={clones} />
+            <InfoPanel data={infoRows} />
+            {clones && <CloneTable dataSource={clones} />}
```
meganrm, line 112: this could also be an optional prop called `subTable` or `additionalInfo` that takes a component, so this component doesn't need to know about clones

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
interim17: sure thing

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -1,81 +1,80 @@
...
+export interface NormalProps extends BaseCellLineProps {
+    orderPlasmid: string;
+    cloneNumber: number;
+    taggedGene: { name: string; symbol: string; structure?: string; protein?: string }[];
+    protein: string[];
+    fluorescentTag: string[];
+    alleleCount: string[];
+}
+
+export interface InfoTableRow {
```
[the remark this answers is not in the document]
interim17: no, removed export

PR #270, opened 2025-07-24
Subpage data: editing design tab

rugeli:
Problem
=======
Part of #254 

This continues the cell line data migration work, following the image data migration in #265 

Solution
========
- migrated existing editing design data for cell lines 
- added video and its thumbnail display
- twisted privacy settings in AICS Vimeo account to get videos to render  
- added schema for admin page data entry  

## Type of change

Steps to Verify:
----------------
- [x] add link to preview [here ](https://deploy-preview-270--cell-catalog.netlify.app)
- [x] add link to admin preview [here](https://deploy-preview-270--cell-catalog.netlify.app/admin) (and verify it loads)

### DIFF
--- a/src/pages/cell-line/AICS-11-27/index.md
+++ b/src/pages/cell-line/AICS-11-27/index.md
@@ -25,13 +25,14 @@ images_and_videos:
     - video: https://player.vimeo.com/video/198395370
       caption: "Timelapse movie of a hiPSC colony expressing tom20-mEGFP. Images taken as a partial z-stack near the bottom  of the cell every 11 seconds for 3 minutes on a spinning-disk confocal microscope. Image is a maximum intensity projection. Movie sped up 55x over real time."
 editing_design:
-  TOMM20_ncbi_isoforms: NM_014765
+  ncbi_isoforms:
+    - NM_014765
   crna: AATTGTAAGTGCTCAGAGCT
   linker: GGSGDPPVAT
   cas9: Wildtype spCas9
   diagrams:
-  - title: "mEGFP Insert"
-    images:
-      - image: EditingDesign_gene_figure.png
-        caption: "Top: TOMM20 locus; Bottom: Zoom in on mEGFP insertion site at TOMM20 C-terminal exon"
+    - title: "mEGFP Insert"
+      images:
+        - image: EditingDesign_gene_figure.png
+          caption: "Top: TOMM20 locus; Bottom: Zoom in on mEGFP insertion site at TOMM20 C-terminal exon"
 ---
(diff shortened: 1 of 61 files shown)


interim17 (commented):
Thanks for getting this going! Don't want to set you back, but I'm slightly concerned about the number of unique fields this is going to generate.

src/pages/cell-line/AICS-11-27/index.md
```diff
@@ -24,4 +24,14 @@ images_and_videos:
       caption: "Z-stack of live hiPSC cells expressing Tom20-mEGFP imaged on a spinning-disk confocal microscope. Image starts at the bottom of the cells and ends at the top."
     - video: https://player.vimeo.com/video/198395370
       caption: "Timelapse movie of a hiPSC colony expressing tom20-mEGFP. Images taken as a partial z-stack near the bottom  of the cell every 11 seconds for 3 minutes on a spinning-disk confocal microscope. Image is a maximum intensity projection. Movie sped up 55x over real time."
+editing_design:
+  TOMM20_ncbi_isoforms: NM_014765
```
interim17, line 28: Do we want to make this more general so that we don't have to have admin and query fields for each isoform type?
Something like:
```
ncbi_isoforms:
- gene: TOMM20
- isoform: NM_014765
```
and then query is generalized:
```
editing_design {
   ncbi_isoforms {
     gene
     isoform
   }
}

Comment (interim17) [reply src/pages/cell-line/AICS-11-27/index.md:28]: If the migration from the old repo would be a lot harder this way I'm happy to work on a script in this repo that spits the fields up.

Comment (rugeli) [reply src/pages/cell-line/AICS-11-27/index.md:28]: oh yeah, we should definitely generalize the field names. I'll make the changes. I'll also check with Megan about moving the giant json file here to make data fetching easier.

Comment (rugeli) [reply src/pages/cell-line/AICS-11-27/index.md:28]: Actually, before putting more work into this field, I'm going to double-check with the editing team. There's a chance we might end up omitting it and just pulling associated isoforms from the gene symbol.

**Comment by rugeli** [comment]
7/28 notes:
- In AICS-0023 data, `"EditingDesign_NCBI_isoform": "NM_003257.4",` is not a typo, so the value doesn't always match the data in 
```
"Main_isoforms": [
      {
        "gene_name": "TJP1",
        "gene_isoforms": [ "NM_003257", "NM_175610" ]
      }
    ],
```
- Brock suggested using genomic regions instead of isoforms in editing design tab. e.g. https://www.ncbi.nlm.nih.gov/projects/sviewer/?id=NC_000015.10&from=29699367&to=29969049 (although I couldn't find this data stored anywhere in the 1.0 json file)
- (not sure if I understood it correctly) Brock also suggested we consider using something programmatic instead of static diagrams to represent gene editing data.

interim17 (approved):

meganrm (approved):

PR #272, opened 2025-08-01
add editing design subpage

interim17:
Problem
=======
Advances #254 
Adds subpage component, and data for editing design to normal cell line subpage.

Solution
========
Made a new unpacking function and type `UnpackedNormalCellLineFull` that follows the pattern used in the disease lines.

Feels a little bit not DRY the way the unpacking functions are so similar (`unpackNormalFrontMatterForSubpage` and `convertFrontmatterToNormalCellLines`) but this data (editing design, etc.) data isn't needed in the table where we use the other unpacking function, so sticking with the pattern.

Fixed what I think was a typo in the `config.yml` crRNA -> crna which is what the pages each have for that field.


src/components/SubPage/EditingDesign.tsx
```diff
@@ -36,9 +36,36 @@ const EditingDesignSubpage: React.FC<EditingDesignSubpageProps> = ({
...
 }) => {
     const rows: DescriptionsProps["items"] = [];
 
+    if (ncbi_isoforms) {
+        rows.push({
+            key: "isoforms",
+            label: "NCBI Isoforms:",
+            children: ncbi_isoforms,
+        });
+    }
```
interim17, line 51: All the fields on these props were optional, so I just added more optional fields and conditionals here rather than make a separate normal/disease editing panel component.

It isn't type-safe, but as long as the fields for normal/disease lines are distinctly named it shouldn't be an issue, only possible confusion or conflict is `crna` and `crnaTargetSite` where the disease lines includes bracketed mutation sites. I'm ok to rely on the queries here to keep these honest, but can do a refactor if we want these two types of editing design data explicitly typed.

src/components/SubPage/convert-data.ts
```diff
@@ -141,6 +149,51 @@ export const unpackParentLineFromFrontMatter = (data: ParentalLineFrontmatter):
     }
 }
 
+export const unpackNormalFrontMatterForSubpage = (
+    cellLineNode: NormalCellLineNode
+): UnpackedNormalCellLineFull => {
```
interim17, line 154: This function largely duplicates `convertFrontmatterToNormalCellLines`, and I think the interfaces for `UnpackedNormalCellLine`, `UnpackedNormalCellLineFull`, `UnpackedDiseaseCellLine` and `UnpackedDiseaseCellLineFull` are not the best, but for time's sake I just followed the existing pattern and prioritized not adding editing design data to a query where it would be unpacked, but then unused, in other components.

meganrm: for context the non-"Full" ones are for the table, which don't request all the data, and the "Full" ones are for the subpages

meganrm (approved):

rugeli (approved):
Looks good except `crRNA` is the correct term. We might want to update all `crna` to `cr_rna` in a separate pr, and stick with crRNA on the site.

static/admin/config.yml
```diff
@@ -221,8 +221,8 @@ collections:
                             hint: "Use / to separate multiple isoforms (e.g., NM_002859 / NM_001080855)"
                         },
                         {
-                            label: "crRNA",
-                            name: "crRNA",
+                            label: "cRNA",
+                            name: "crna",
```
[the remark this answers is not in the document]
interim17: I just changed it make it match what all the files in `/pages/cell-line/*` have:
```
editing_design:
  ncbi_isoforms:
    - NM_014765
  crna: AATTGTAAGTGCTCAGAGCT
  linker: GGSGDPPVAT
  cas9: Wildtype spCas9
  ```
 
 I can change this label, but we want `crrna` in the `name` field we will need to adjust all those markdown files.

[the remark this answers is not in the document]
interim17: Should the label be `cr RNA` or `crRNA`?

PR #274, opened 2025-08-02
Feature/lint

interim17:
Problem
=======
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

Solution
========

with @pairperson1

## Type of change

* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* This change requires a documentation update
* This change requires updated or new tests

Steps to Verify:
----------------
- [ ] add link to preview here 
- [ ] add link to admin preview here (and verify it loads)  

1. A setup step / beginning state
2. What to do next
3. Any other instructions
4. Expected behavior
5. Suggestions for testing

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

Keyfiles (delete if not relevant):
-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!


gatsby-config.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
     //     DEV_SSR: true,
     // },
     plugins: [
-        "gatsby-plugin-react-helmet",
+        "gatsby-plugin-react-helmet-async",
```
interim17, line 24: It seems like for SSR the async version of react-helmet is preferred, and is also more maintained. I have yet to confirm what it does, and what this has changed. I just made the linter happy.

The Helmet is an example of something where I am unclear if we are intentionally implementing it or it is a legacy from the template used to start the repo.

src/cms/preview-templates/DiseaseCatalogPreview.tsx
```diff
@@ -2,13 +2,14 @@ import React from 'react'
...
+        title: entry.getIn(["data", "title"]),
+        content: widgetFor("body"),
+        footerText: entry.getIn(["data", "footer_text"]),
+    };
 
+    return <DiseaseCatalogTemplate {...(props as any)} />;
+};
 
-export default DiseaseCatalogPreview;
+export default DiseaseCatalogPreview;
```
interim17, line 15: This is a punt, along with the override rules in the eslint config, the preview code is too much to deal with right now.

src/component-queries/DiseaseCellLines.tsx
```diff
@@ -51,15 +51,15 @@ const DiseaseCellLinesTemplate = (props: DiseaseCellLinesTemplateProps) => {
     const { edges: cellLines } = props.data.allMarkdownRemark;
     const { diseases } = props;
     const groupedCellLines = groupLines(diseases, cellLines);
+    const width = useWindowWidth();
+    const isPhone = width < PHONE_BREAKPOINT;
+
+    const suppressRowClickRef = useRef(false);
```
interim17, line 57: needed to pull hooks out of `map`

src/components/CellLineInfoCard/CellLineInfoCardBase.tsx
```diff
@@ -54,13 +51,18 @@ const CellLineInfoCardBase = ({
...
+                            await navigator.clipboard.writeText(href);
                             setShareTooltipText("Copied!");
+                        } catch {
+                            setShareTooltipText("Failed to copy");
+                        } finally {
                             setTimeout(() => {
                                 setShareTooltipText("");
                             }, 1000);
-                        });
+                        }
```
interim17, line 65: `navigator.clipboard.writeText(href)` is async. I made one linting rule happy by adding this try/catch, but added an ignore to the other one for now.

src/components/CellLineTable/NormalTableColumns.tsx
```diff
@@ -73,8 +74,8 @@ const obtainPlasmidColumn = {
 
 export const getNormalTableColumns = (
     inProgress: boolean
-): CellLineColumns<UnpackedNormalCellLine> => {
-    const columns = [
+): ColumnsType<any> => {
+    const columns: ColumnsType<any> = [
```
interim17, line 78: Tweaking the column typing was one of the more involved changes.

I typed the columns prop in `CellLineTable` and I'm using the `antd` type `ColumnType` and `ColumnProps` to get the same type-safety we got from this before:
`export type CellLineColumns<T> = GetProp<typeof Table<T>, "columns">`

Probably more than one way to do this, but everything seems to be working...

src/components/Layout.tsx
```diff
@@ -13,66 +13,74 @@ const TemplateWrapper = ({ children }: React.PropsWithChildren) => {
     const { title, description } = useSiteMetadata();
     return (
         <ConfigProvider theme={theme}>
-            <Helmet>
-                <html lang="en" />
-                <title>{title}</title>
-                <meta name="description" content={description} />
+            <HelmetProvider>
```
interim17, line 16: Wrapping in `HelmetProvider` seems to be the main difference here. As I said in the write up I'm a little shaky on exactly how this all works, would appreciate feedback.

src/components/Navbar.tsx
```diff
@@ -13,9 +11,7 @@ const Navbar = () => {
       <div className="container">
         <ul
           id="navMenu"
-          className={` navbar-start has-text-centered navbar-menu ${
-            isActive && "is-active"
-          }`}
+          className={` navbar-start has-text-centered navbar-menu`}
```
interim17, line 14: `isActive` was never reset, so that classname was never applied. Also is this component going to be used? Doesn't seem to be currently.

src/hooks/useEnv.ts
```diff
@@ -15,7 +15,7 @@ const useEnv = () => {
             // to the staging environment
             setEnv("staging");
         }
-    }, []);
+    }, [hostname]);
```
interim17, line 18: Linter didn't like missing dependency in hook

src/pages/404.js
```diff
@@ -2,12 +2,12 @@ import * as React from "react";
...
-      <h1>NOT FOUND</h1>
-      <p>You just hit a route that doesn&#39;t exist... the sadness.</p>
-    </div>
-  </Layout>
+    <Layout>
+        <div>
+            <h1>NOT FOUND</h1>
+            <p>You just hit a route that doesn&#39;t exist... the sadness.</p>
+        </div>
+    </Layout>
```
interim17, line 10: Do we want this?

src/templates/disease-catalog.tsx
```diff
@@ -35,7 +35,7 @@ interface DiseaseCatalogTemplateProps {
     coriellImage: FileNode;
     coriellLink: string;
 }
-// eslint-disable-next-line
+
```
interim17, line 38: This was a little confusing, we have these `eslint` comments a few places, but were not running `eslint`... template legacy? They don't seem necessary with the current config.

meganrm: I have eslint sent to run in my IDE, so that's probably what the origin of these are

Issue #275, opened 2025-08-04
EPIC: Linting, CI/CD, formatting, testing


interim17:
## Use Case

We could improve pace of development and have more reliable build and deployment with the use of a linter, pre-commit checks, defined formatting rules, and possibly testing.

## Acceptance Criteria

Sub tasks

- [ ] Install `eslint`, define script (minimal rules at first?)
- [ ] Define typecheck script
- [ ] Install `prettier`, define scripts for formatting check and fix
- [ ] Use linting, formatting and typechecking in CI/CD flow
- [ ] Install `husky` for pre-commit checks
- [ ] Run formatting script on repo and commit changes
- [ ] Overtime: add more linting rules and adjust repo to conform
- [ ] If desired: add testing framework
- [ ] If desired: explicit typing on graphql schema

## Details

_Please provide any helpful specifications_

meganrm:
@interim17 maybe break this up? I feel like it would be nice to close this as an MVP but a few of the things could be captured in a new issue

PR #278, opened 2025-08-04
Feature/typecheck

interim17:
Problem
=======
Advances #275 

Solution
========
Add typecheck script and fix the things that were flagged.

`react-helmet` to `react-helmet-async` 
Typing wasn't working for `react-helmet` but we are on React 18 and using SSR so we should be using the async version anywhere according to what I found online.
Wrapped the Layout in `HelmetProvider` and installed `react-helmet-async` and `gatsby-plugin-react-helmet-async`
Wasn't paying close attention to how this worked, but from what I can tell the site metadata is still there on my local instance.

I don't know how best to handle the DiseaseCatalogPreview component, and the fact that we don't pass it the full set of things it requires for props. I typecast the props as `any` to bypass type checking, it's hack... Open to ideas.

`MobileView` was treating array props as individual entries so I mapped them.

Uncommented this line in ts config: `"types": [],` so that downstream packages wouldn't through type check errors.


src/component-queries/types.ts
```diff
@@ -32,6 +32,12 @@ export interface ParentalLineFrontmatter {
     allele_count: string[];
     tag_location: string[];
     fluorescent_tag: string[];
+    images_and_videos?: {
+        images: {
+            image: FileNode;
+            caption: string;
+        }[];
+    };
```
interim17, line 40: There are inflight changes here, will follow along and update accordingly...

PR #277, opened 2025-08-04
testing CI/CD

interim17:
Problem
=======
What is the problem this work solves, including
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

Solution
========
What I/we did to solve this problem

with @pairperson1

## Type of change
Please delete options that are not relevant.

* Bug fix (non-breaking change which fixes an issue)
* New feature (non-breaking change which adds functionality)
* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* This change requires a documentation update
* This change requires updated or new tests

Steps to Verify:
----------------
- [ ] add link to preview here 
- [ ] add link to admin preview here (and verify it loads)  

1. A setup step / beginning state
2. What to do next
3. Any other instructions
4. Expected behavior
5. Suggestions for testing

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

Keyfiles (delete if not relevant):
-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

PR #279, opened 2025-08-05
Fix/decap cms version

rugeli:
Problem
=======
patch to #277 to solve decap cms admin page loading issue

Solution
========
- located the incompatible plugin `gatsby-transformer-remark-frontmatter` which requires gatsby 3.x but we have gatsby 5
- pinned decap-cms-app to use compatible version 

pro: this way we get to keep node20
con: the @md extension has been removed with the incompatible `gatsby-transformer-remark-frontmatter`, might want to look for another approach to convert frontmatter to html 

## Type of change

Steps to Verify:
----------------
- [x] add link to preview [here ](https://deploy-preview-279--cell-catalog.netlify.app/)
- [x] add link to admin preview [here ](https://deploy-preview-279--cell-catalog.netlify.app/admin)(and verify it loads)  

1. rm -rf node_modules yarn.lock && yarn install


interim17 (approved):
I think this is fine as it is, but would be more complete if we merge:
https://github.com/allen-cell-animated/cell-catalog/pull/281/files
into this branch first!

The types are fixed here, but the formatting we got for free from the html-in-markdown was lost.

PR #287, opened 2025-08-12
custom icons for tube and plasmid

interim17:
Problem
=======
Advances #254 

We need to add these icons to the normal cell line info cards, I wanted to pull these components out to make the JSX and CSS a little more tidy. See [next PR for context.](https://github.com/allen-cell-animated/cell-catalog/pull/288)


src/style/table.module.css
```diff
@@ -90,19 +90,23 @@
     color: var(--WHITE);
 }
 
-.action-column:hover .action-button .tube-icon path,
-.action-column:hover .action-button .tube-icon line {
+.action-column:hover .action-button :global(.tube-icon) path,
+.action-column:hover .action-button :global(.tube-icon) line {
```
interim17, line 94: Global class name instead of importing unique class names anywhere we need to style the icons.

meganrm (approved):
agree with ruge on the double type

interim17:
> agree with ruge on the double type

Will remove!

PR #293, opened 2025-08-19
consolute styling across catalogs

meganrm:
Problem
=======
the styles were still inconsistent between the main catalog and the disease catalog, and they were being styling in two different places. There were also more divs and styles than needed for the layout

Solution
========
combined the `about` styling into one place


src/style/about.module.css
```diff
@@ -1,30 +1,27 @@
...
+.content-wrapper {
+    margin-bottom: 16px;
 }
 
-.section h1 {
-    font-size: 33px;
-}
-
-.first-block {
+.content-wrapper :global(.content) {
```
interim17, line 10: Curious the thinking here with the global class name. I see it being applied on a few components, but this is the only place I'm seeing it being referenced in a selector.

interim17 (approved):
Looks good, appreciate the consolidation. Left one question about the global content class name, non-blocking.

PR #300, opened 2025-08-21
genomic characterization data

rugeli:
Problem
=======
subpage data migration and update to the admin data schema 
advances #254 

Solution
========
- transformed and migrated existing genomic characterization data from catalog 1.0
- updated the admin data page to support genomic characterization data entry

Note: The data structure may require further updates once the design is finalized 

## Type of change

Steps to Verify:
----------------
- preview [here ](https://deploy-preview-300--cell-catalog.netlify.app/)
- admin preview [here ](https://deploy-preview-300--cell-catalog.netlify.app/admin)


meganrm (approved):

interim17:
I wanted to mention that I've started working with this data, and it seems like some of the file paths for the images are present in the folders and some are not.

This is leading to errors so I think we should address two things:
- make sure all the files referenced in this data exist
- but also look at some of our data un packing functions and address that they won't throw un caught errors if certain fields are falsy

For example `GenomicCharacterization_junction_schematic_mEGFP.png` doesn't exist for line 16, so trying to access imageObj.image.childImageSharp throws an error in `unpackMultiImageDiagrams`

If we merge this branch we won't have errors because we don't query this data anywhere, so its not the end of the world to merge this PR as is.

rugeli:
Follow-up ticket #308: forward junction schematic pngs to this repo, and cleanup duplicate images.

PR #305, opened 2025-08-26
Feature/anchor alias

interim17:
Problem
=======
This is not a high-priority PR, but one that I like. Adding anchor/alias pairs to places in the config yaml where we are repeating data structures.

I like this for this repo, as it is more concise, but also seems like a good pattern if we want to keep using this sort of CMS for AICS tools. I think we could go further and compose with js which would allow for the use of actual constants, and might integrate better with unit testing? But this is a start. 

Curious if reviewers think more comments or documentation would be helpful or if the pattern is fairly self explanatory?


static/admin/config.yml
```diff
@@ -106,61 +174,15 @@ collections:
...
-                label: "Certificate of Analysis",
-                name: "certificate_of_analysis",
-                widget: "file",
-                required: false,
+                fields: [
+                { <<: *relation_gene, required: false },
+                { label: "Allele Count", name: "allele_count", widget: "select", options: *options_alleleCount, required: false },
+                { label: "Tag Location", name: "tag_location", widget: "select", options: *options_tagLocation, required: false },
+                { label: "Fluorescent Tag", name: "fluorescent_tag", widget: "select", options: *options_fluorescentTag, required: false }
+                ]
```
interim17, line 182: In a case like this the gene relation has some basic fields that are always present, but some that we add on a case-by-case basis

PR #306, opened 2025-08-29
use camel case and alias fields in config and markdown for iamges

interim17:
Problem
=======
Some tweaks to the genomic characterization data and config as prep for the sub tab work.

3 changes:
- `editedGene` to `edited_gene` in config/markdown to match our usual pattern
- schema customization: some of the image files are missing, so without schema customization gatsby defaults the file paths to type `string` and then the queried subfields cause errors
- anchor/alias pattern in yml to make it clear what is when when dealing with images/videos/diagrams

to do: we have mostly multi image diagrams, and one remaining single image diagram in the config (disease genomic characterization). Maybe we should consolidate around the multi image diagram data structure?

The subpage work has enough code I think this prep work should be reviewed separately.


meganrm:
I'm waiting on your to resolve conflicts before looking at this

meganrm (approved):

PR #307, opened 2025-08-29
genomic characterization tab

interim17:
Problem
=======
Advances #311

Solution
========
Essentially finishes the feature, although it seems there is still missing data and a few other things to clean up before release.

Feedback from UX was to use a masonry/mosaic pattern with all tables at 50% of container with and equal margins between tables and diagrams. Will repeat this pattern for stem cell characteristics.

A little bit of data cleanup necessary to prevent errors is included here, using `null` in place of the string "N/A" which appeared in some fields that expected float data.

Screenshot 2025-09-11 at 2 26 43 PM
Screenshot 2025-09-11 at 2 26 38 PM


src/components/SubPage/GenomicCharacterization.tsx
```diff
@@ -1,35 +1,66 @@
...
+import { UnpackedGenomicCharacterization as GenomicCharacterizationProps } from "./types";
+import SubpageTable from "../shared/SubpageTable";
+import { DDPCR_COLUMNS, CRRNA_OFF_TARGETS_COLUMNS, AMPLIFIED_JUNCTION_COLUMNS } from "./genomic-characterization-table-constants";
 
 const {
-    card,
-    fullWidth,
+    masonry,
+    masonryItem,
+    masonrySpanAll,
```
meganrm, line 11: because these are helper classes I would put them in a higher level module (do we have a general styles.css)?

src/components/SubPage/GenomicCharacterization.tsx
```diff
@@ -1,35 +1,66 @@
...
+                <SubpageTable
+                    className={masonryItem}
+                    title={"GFP and donor plasmid copy number"}
+                    columns={DDPCR_COLUMNS}
+                    dataSource={ddpcr.data}
+                    caption={ddpcr.caption}
+                />
+            ) : null}
+            {crRnaOffTargets?.data.length ? (
+                <SubpageTable
```
meganrm, line 42: can this subpage table just handle checking for the length of data? so this file doesn't have to have multiple  conditionals. I like that it's either an empty list or a data list, I don't like components having to check if data is of two different types, ie Null or a List, but I think handling an empty list seems ok

src/components/SubPage/GenomicCharacterization.tsx
```diff
@@ -1,35 +1,66 @@
...
+            ) : null}
             {diagrams?.map((diagram, index) => (
                 <DiagramCard
                     key={index}
                     className={classNames(
-                        card,
-                        diagram.title?.includes("Sanger") ? fullWidth : ""
+                        diagram.title?.includes("Sanger")
+                            ? masonrySpanAll
+                            : masonryItem
```
meganrm, line 56: nit but you don't need classNames here anymore

src/style/genomic-characterization.module.css
```diff
@@ -1,19 +1,34 @@
...
+.masonry {
+    column-count: 2;
+    column-gap: 40px;
+    width: 100%;
 }
 
-.card.full-width {
-    flex-basis: 100%;
-    max-width: 100%;
+.masonry :global(.ant-card).masonry-item {
```
meganrm, line 7: I don't like that this switched from not needing a global tag to using one. was .masonary-item not sufficient? can you just keep the .card class instead?

interim17: I didn't want it that way either, but there is a media query on the global class that was over riding it, it's not necessary for the base clas per se, but it is necessary for the media query and I wanted them to have the same selector

src/components/SubPage/GenomicCharacterization.tsx
interim17: I'm going to move these rules into `subpage.module.css` since that seems to be a higher level but also specific to the components that will use these rules

src/components/SubPage/SubpageTabs.tsx
```diff
@@ -33,20 +35,32 @@ const SubpageTabs: React.FC<SubpageTabsProps> = ({
             </div>
         );
     };
+
+    const hasGcTables = (gc?: UnpackedGenomicCharacterization | null): boolean =>
+        !!(
+            gc?.amplifiedJunctions?.data?.length ||
+            gc?.ddpcr?.data?.length ||
+            gc?.crRnaOffTargets?.data?.length
+        );
```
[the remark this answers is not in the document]
interim17: So I adjusted the way this is done:
- used two small utils at the `convert-data` level to ensure that the property is left undefined if the array is empty
- ditched the checks in this component, since each piece of data gets its own conditional check anyway down the line
- added a null check on the table component

The only risk here is that if the GC object is defined, but the properties are all undefined, we wont get our "no data component", but I'm not sure if we really want that component in production anyway, or if it was more of a dev placeholder?

src/components/SubPage/convert-data.ts
```diff
@@ -92,6 +95,56 @@ export const unpackEditingDesignData = (editing_design?: {
...
+                totalSitesSequenced: offTarget.total_sites_sequenced,
+                mutationsIdentified: offTarget.mutations_identified
+            }
+        })
+    }
+    const data = {
+        amplifiedJunctions: hasTableData(amplifiedJunctionsData) ? amplifiedJunctionsData : undefined,
+        ddpcr: hasTableData(ddpcrData) ? ddpcrData : undefined,
+        crRnaOffTargets: hasTableData(crRnaOffTargetsData) ? crRnaOffTargetsData : undefined,
+        diagrams: nonEmptyArray(diagrams) ? diagrams : undefined,
```
meganrm, line 141: take it or leave it, but I would maybe make a default `emptySubpageData` object that has all the keys you need, `data: [], caption: ""` and then fill it in if the data is there. That way you don't have as many `if exists` checks.

meganrm (approved):
looks good, I added a few comments on a different way to approach null values. I think it's nice to have things defined but empty

PR #309, opened 2025-09-02
add typecheck script

interim17:
Problem
=======
Advances #275 

Solution
========
Add a `typecheck` script to `package.json` and cleared up current issues flagged when running the script.

Some of the fixes are patch/hacky. In the interest of simplicity I tried to keep this change set as minimal as possible. I didn't adjust the `react-helmet` dependency to use `react-helmet-async` or change the type packages, or anything in ts.config. Just aiming to provide typecheck script that can be easily used before committing changes.

- add script
- add `global.d.ts` file to declare decap media libraries
- adjust prop typing for `Helmet`, icons, and preview
- in mobile view, map over tagged gene since it is a list.

Once we have scripts for linting, formatting, etc all cleaned up we can work on CI/CD integration, this is optional for now.


src/global.d.ts
```diff
@@ -0,0 +1,3 @@
+// Decap CMS Media Libraries
+declare module "decap-cms-media-library-uploadcare";
+declare module "decap-cms-media-library-cloudinary";
```
meganrm, line 3: are there not types we can install instead?

interim17: Not that I found

src/components/Layout.tsx
```diff
@@ -9,6 +9,10 @@ import useSiteMetadata from "./SiteMetadata";
 import theme from "../style/theme";
 const { container } = require("../style/layout.module.css");
 
+const Helmet = RawHelmet as unknown as React.ComponentType<
+    React.PropsWithChildren<any>
+>;
```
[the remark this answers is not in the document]
interim17: Yes, that's what I want to do long term. But because we ran into so many problems last time with PRs where we changed dependencies, and because this is a low priority ticket, I was trying to do this without any dependency changes.

rugeli (approved):

PR #340, opened 2025-09-29
Feature/disable buttons

interim17:
Problem
=======
Mostly this adds disabled styling to table and subpage buttons when the order links are missing.

Although I did notice that the order link seemed to be missing in the cell-line query so that's a crucial add!

Screenshot 2025-09-29 at 1 13 56 PM
Screenshot 2025-09-29 at 1 13 50 PM


src/components/CellLineTable/SharedColumns.tsx
```diff
@@ -94,21 +96,29 @@ export const obtainLineColumn = {
...
-                </a>
-            )
+        const isDisabled = !orderLink;
+        const link = (
+            <a
+                key={orderLink}
+                className={classNames(actionButton, {
+                    [disabled]: isDisabled,
+                })}
+                href={isDisabled ? undefined : orderLink}
```
meganrm, line 106: does an empty string still make it clickable? just curious

src/components/CellLineInfoCard/NormalCellLineInfoCard.tsx
```diff
@@ -87,12 +87,14 @@ export const NormalCellLineInfoCard: React.FC<NormalCellLineInfoCardProps> = (
         {
             key: "order",
             label: `Obtain ${formatCellLineId(props.cellLineId)}`,
+            disabledLabel: `Collection not yet available.`,
```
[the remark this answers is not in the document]
interim17: Like above the feedback was actually to make them match the button label so I'm going to change this to:
```
disabledLabel: `${formatCellLineId(
                props.cellLineId
            )} not yet available.`,

src/components/CellLineTable/SharedColumns.tsx
meganrm: so could it just be `href={orderLink}` ? since isDisabled is just checking if it exists?

rugeli (approved):

PR #377, opened 2025-12-04
Feature/nav bar

### LINKED ISSUE #372: nav bar like CFE, TFE, simularium
## Use Case

we need a nav bar that says the name of the site and links to the main allencell.org site

## Acceptance Criteria

question: should it be fixed, leaning towards no

## Details

branch from `feature/banner`

interim17:
Problem
=======
Closes #372

Solution
========
Updated 12/9/25:
- NavBar exists as a component query with a markdown file that is not currently in the admin/config but can be added
- added `NavBarDropdown` component with a bunch of little handlers for shaping the data and getting the items props with CSS classnames for the antd component to consume
- NavBar dropdowns are populated and styled for Catalogs, Protocols, and Collections
-  added `/static/pdf` dir and new section in `gatsby-config.js` using `gatsby-source-filesystem` to make those pdfs accessible as files
- uploaded protocol pdfs from the old cell catalog site

Possible TODOs:
- styling could be tweaked on the `Collections` menu items its not pixel perfect yet.

Screenshot 2025-12-09 at 10 08 31 AM
Screenshot 2025-12-09 at 10 08 26 AM
Screenshot 2025-12-09 at 10 08 21 AM
Screenshot 2025-12-09 at 10 08 13 AM

### DIFF
--- a/src/component-queries/NavBar.tsx
+++ b/src/component-queries/NavBar.tsx
@@ -12,7 +12,7 @@ const {
     divider,
     leftContent,
     logoLink,
-    pageHeader,
+    container,
     rightContent,
     titleLink,
 } = require("../style/navbar.module.css");
@@ -73,7 +73,7 @@ const NavBar: React.FC = () => {
     );
 
     return (
-        <div className={pageHeader}>
+        <div className={container}>
             <div className={content}>
                 <div className={leftContent}>
                     <a
--- a/src/style/navbar.module.css
+++ b/src/style/navbar.module.css
@@ -1,4 +1,4 @@
-.page-header {
+.container {
     background-color: var(--SERIOUS_GRAY);
     height: 61px;
     padding: 13px 20px;


src/style/navbar.module.css
```diff
@@ -0,0 +1,37 @@
+.page-header {
```
meganrm: nit: but we usually call the highest level class just `container`. I'm guessing this is just what we called it in cfe? Alternatively `nav-bar` would be more consistent with the component name

meganrm:
the hover state on the links out to correl end up having a little dot on the end, because the underline doesn't underline the link out icon but does underline the parentheses. A simple solution is to just move the linkout icon out of the parens because trying to have the underline just stop at the words seems like a pain

src/component-queries/NavBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    <span className={divider}>|</span>
+                    <NavBarDropdown
+                        buttonComponent={
+                            <div className={titleLink}>
+                                Allen Cell Collection
+                            </div>
+                        }
+                        label="Catalogs"
+                        items={catalogs}
+                    />
```
meganrm: looking more closely at the design, since so much of this is already hardcoded it seems like having this single component that has to do so much formatting is overly general. I would map over catalogs here and send them in as already formatted menu items (that don't open a new tab)

interim17:
@meganrm Thanks for all the feedback!

Did a big re-org and a number of small fixes to bugs and styling.

Also sent Travis a couple questions about the catalog items.

Made one change for my sanity that I think you might hate and added an anchorType field to the markdown to make it simple to know at a glance if something is an internal/external link or a pdf download. It's just much easier for me mentally and makes downstream functions more readable.

Changes:
- NavBarDropdown directory exists with a `constants.ts` for the catalog component stuff, and a `formatDropDownMenuItems` file to handle formatting, which is called at the `NavBar` level
(Question, should I move NavBar into this folder? I left it in component queries because that what it is... I could put the directory there?)
- `PdfAnchor` and `LinkOut` components with prop to control new tab behavior
- made href field required to reduce checks
- added styling to triggers when dropdowns are open
- bumped icon outside the parens on collection links

src/components/LinkOut.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+interface LinkOutProps {
+    label: string | React.ReactNode;
+    href: string;
+    newTab?: boolean;
+}
+
+export const LinkOut = ({label, href, newTab}: LinkOutProps) => {
```
meganrm: nit: since this is called LinkOut, and you have an optional prop, I would think the default behavior is to have it go out. 

Also, right now the right hand links are not linking out, and I'm pretty certain they're supposed to be because they have the external icon (sorry if I put my comment about the left hand side not linking out in the wrong line and you changed it because of my comment). It can also be confusing to have an optional boolean because you don't know you're making a choice by not sending a prop in 
 
my suggestions: 
1. call this LinkOut, and have an optional `samePage` disable the default `target=_blank` behavior 
2. have this just called Link, and make a required prop that's an enum of either `SamePage` `NewTab` so you have to explicitly choose

src/components/NavBarDropdown/formatDropDownMenuItems.tsx
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const getItemComponent = (item: NavBarDropdownItem) => {
+    if (item.anchorType === NavBarAnchorType.Download) {
+        const filename = item.href.split('/').pop();
+        return <PdfAnchor label={item.label} href={`/pdf/${filename}`} />;
+    }
+
+    if (item.anchorType === NavBarAnchorType.Internal) {
+        return <LinkOut label={item.label} href={item.href}/>;
```
meganrm: this is where link out looks confusing, i didn't notice the `newTab=true` below so it seemed weird to have Internal => LinkOut

src/components/NavBarDropdown/formatDropDownMenuItems.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    if (COLLECTION_COMPONENTS_MAP[item.label]) {
+        return <LinkOut label={COLLECTION_COMPONENTS_MAP[item.label]} href={item.href} />;
```
meganrm: this needs `newTab={true}`

rugeli (approved):
clicked around the nav bar and everything worked as expected! 
One small thing I noticed on my end: I don’t see the parenthesis text in that menu, but it doesn’t seem to affect functionality. All good otherwise!
Screenshot 2025-12-10 at 3 59 09 PM

interim17:
> clicked around the nav bar and everything worked as expected! One small thing I noticed on my end: I don’t see the parenthesis text in that menu, but it doesn’t seem to affect functionality. All good otherwise! Screenshot 2025-12-10 at 3 59 09 PM

I talked to Travis about this and we tweaked this design on the fly!

PR #386, opened 2025-12-16
delete defaultProps from imported SVGs

### LINKED ISSUE #381: Get rid of `defaultProps` warnings
## What needs to happen?
we shouldn't have the warnings in the browser console. They are caused by the svg elements being imported into react

interim17:
Problem
=======
Closes #381 

Solution
========
Got `gatsby-plugin-svgr-svgo` configured to drop-in replace `gatsby-plugin-react-svg` and remove defaultProps warnings.

~~This is a bit of a hack, just typecasting and deleting the defaultProps that seem to be added by `gatsby-plugin-react-svg`~~

~~Likely the better long-term fix is to move away from that plugin:
https://github.com/jacobmischka/gatsby-plugin-react-svg/issues/59~~

~~My first few attempts to switch plugins were met with build issues. If we want a quick solution, I think this will remove the browser warnings, and we can look at switching plugins going forward?~~

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -170,10 +170,24 @@
     js-tokens "^4.0.0"
     picocolors "^1.0.0"
 
+"@babel/code-frame@^7.28.6":
+  version "7.28.6"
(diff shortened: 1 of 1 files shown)


interim17:
@meganrm if you want to prioritize speed over beauty we can take this approach as a stop gap?

PR #387, opened 2025-12-16
Fix/static queries

### LINKED ISSUE #382: Clean up staticQuery warnings
## What needs to happen?

update the static query components

## Why should we do this?
it's deprecated 

## When does this need to get done?
not an emergency but nice to have done

interim17:
Problem
=======
Closes #382 

Solution
========
Refactor all the uses of `<StaticQuery />` components to use the `useStaticQuery` hook to query data and then just return plain components.

This gets rid of the warnings in my browser.

There is a lot of line movement here in the refactor but should be no functional changes, just shifting things around. `NavBar` was the proof of concept for this hook.


rugeli (approved):
LGTM, nice cleanup!

PR #393, opened 2025-12-19
move url formatting into component, and open in tab

meganrm:
Problem
=======
from travis: Each of the protocols should have a download icon next to them (see attached image, 16px size icon). Ideally these appear only on hover. An alternative would be if the PDF could open in a new tab without downloading anything (the DL icon would not be required in this case)

Solution
========
I opted for opening in a new tab. 

## Type of change

Steps to Verify:
----------------
1. open link 
2. in the protocols drop down, clicking any of the protocols should open a pdf in a new tab 
- [x] [add link to preview here ](https://deploy-preview-393--cell-catalog.netlify.app)
- [x] [add link to admin preview here (and verify it loads) ](https://deploy-preview-393--cell-catalog.netlify.app/admin/#/)

### DIFF
--- a/src/components/PdfAnchor.tsx
+++ b/src/components/PdfAnchor.tsx
@@ -6,6 +6,9 @@ interface PdfAnchorProps {
 }
 
 export const PdfAnchor = ({ href, label }: PdfAnchorProps) => {
+    if (!href) {
+        return null;
+    }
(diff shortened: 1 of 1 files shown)


frasercl (approved):

ShrimpCryptid (approved):
Tested, LGTM!

PR #411, opened 2026-01-14
swipe media cards on small screen

interim17:
Problem
=======

Solution
========

Screenshots (optional):
-----------------------
Screenshot 2026-01-13 at 4 09 48 PM
Screenshot 2026-01-13 at 4 09 39 PM

### DIFF
--- a/src/style/banner.module.css
+++ b/src/style/banner.module.css
@@ -42,6 +42,7 @@
     .container {
         height: auto;
         padding: 0;
+        background-attachment: scroll;
     }
     .banner-container {
         padding: 100px 25px 50px;


src/components/ImagesAndVideo/ImagePreviewGroup.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const PreviewGroup = ({
+    currentIndex,
+    imageItems,
+    mediaItems,
+    previewVisible,
+    setCurrentIndex,
+    setPreviewVisible,
+    setSelectedMedia,
+}: ImagePreviewGroupProps) => {
```
interim17: Strictly speaking this component doesn't get re-used so doesn't deserve to be a component, but I was trying to reduce line count and separate concerns in `ImagesAndVideo` it does get rendered in two places, but is the same in both.

src/components/ImagesAndVideo/ImagesAndVideos.tsx
```diff
@@ -0,0 +1,195 @@
...
+} from "../../component-queries/types";
+import {
+    useMobileBreakpoint,
+    useTabletBreakpoint,
+} from "../../hooks/useWidthBreakpoint";
+import { formatCellLineId } from "../../utils";
+import { isImage } from "../../utils/mediaUtils";
+import Thumbnail from "../Thumbnail";
+import { PreviewGroup } from "./ImagePreviewGroup";
+import { MediaCard } from "./MediaCard";
```
interim17: I pulled the preview component, and `MediaCard` (along with `renderMediaItems()` into sepraate files to make this a little less intimidating to read.

src/components/ImagesAndVideo/ImagesAndVideos.tsx
```diff
@@ -0,0 +1,194 @@
...
+            className={header}
+            vertical={isMobile}
+            gap={isTablet ? 8 : 0}
+        >
+            <div className={titleSection}>
+                <h3 className={mainTitle}>{formatCellLineId(cellLineId)}</h3>
+                <span className={subtitle}>
+                    {parentalGeneSymbol
+                        ? `${geneSymbol} in WTC-${fluorescentTag}-${parentalGeneSymbol} (${alleleTag}-allelic tag)`
+                        : `${geneSymbol} in WTC-${fluorescentTag} (${alleleTag}-allelic tag)`}
```
meganrm: I'm guessing this is how we had this before, but it looks like something that could be a formatting function instead of inline like this

src/components/ImagesAndVideo/ImagesAndVideos.tsx
```diff
@@ -0,0 +1,195 @@
...
+                {previewGroup}
+                <div className={mobileCardContainer}>
+                    {mediaItems.map((_item, index) => (
+                        <MediaCard
+                            className={mobileMediaCard}
+                            key={index}
+                            index={index}
+                            title={title}
+                            isMobile={isMobile}
+                            item={mediaItems[index]}
```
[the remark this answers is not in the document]
meganrm: I agree with this

src/components/ImagesAndVideo/ImagesAndVideos.tsx
interim17: Easy to make one, I think it feels a bit odd to pass that many props to a util so I'm going to define the util above the render in this file.

src/components/ImagesAndVideo/ImagesAndVideos.tsx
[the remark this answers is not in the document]
interim17: Idk what I did last time but it worked this time! Pushed this a couple days agio:
https://github.com/AllenCell/cell-catalog/pull/411/commits/ef492c5bbb01aab7fffe82425aa068176fc305cf

PR #420, opened 2026-02-10
create unique id for genes

meganrm:
Problem
=======
a number of genes have the same gene symbol but are visualizing different structures 

Solution
========
I made a unique identifier. I tried other methods of automatically creating a unique ID, but struggled to figure out a good solution that worked for all of them. Since this is probably not going to be an ongoing problem I think it's an ok solution. If we were planning on making a lot of new cell lines, I would want to create checks to make sure the id is actually unique. 

## Type of change

Steps to Verify:
----------------
- [ ] add link to preview here 
- [ ] add link to admin preview here (and verify it loads)


rugeli (approved):
Built locally and played around with the preview site, gene related display and search&filter work well!

PR #421, opened 2026-02-13
Feature/null search

### LINKED ISSUE #266: dead end page for null search results
## Use Case

UX has provided designs for dead ends when search results don't turn anything up:
https://www.figma.com/design/kMIJN6BoOkXIhnVUruIHNT/Website-allencell.org?node-id=5836-9870&t=nBLvVnwDN2dpJbO3-4

## Acceptance Criteria

Categories with no results show the dead end message, and are pushed to the bottom of the order.

## Details

interim17:
Problem
=======
Closes #266 

Solution
========
Sort the categories so empty results move to the bottom.

Define a small block of JSX to pass to the table when data is empty, and also hide the header with the `showHeader` prop.

Defined what I felt were reasonable reductions in margins, padding and font for smaller screen situations, as the designs are for a very wide screen.

Screenshot 2026-02-12 at 4 15 49 PM

Screenshot 2026-02-12 at 4 20 41 PM

### DIFF
--- a/src/components/CategorySections.tsx
+++ b/src/components/CategorySections.tsx
@@ -1,15 +1,12 @@
 import React from "react";
 
-
-
-import { CategoryLabel, UnpackedNormalCellLine } from "../component-queries/types";
+import {
+    CategoryLabel,
+    UnpackedNormalCellLine,
+} from "../component-queries/types";
 import CellLineTable from "./CellLineTable";
 import { getNormalTableMobileConfig } from "./CellLineTable/MobileView";
 import { getNormalTableColumns } from "./CellLineTable/NormalTableColumns";
-
-
-
-
 
 interface CategorySectionsProps {
     selectedCategories: CategoryLabel[];


TravisKroeker (approved):
Looks solid

src/components/CategorySections.tsx
```diff
@@ -58,11 +58,19 @@ const CategorySections: React.FC<CategorySectionsProps> = ({
         [filteredList, selectedCategories],
     );
 
+    // Sort categories: non-empty first, empty last (preserves relative order)
+    const sortedCategories = React.useMemo(() => {
+        const nonEmpty = selectedCategories.filter(
+            (cat) => buckets[cat]?.length,
+        );
+        const empty = selectedCategories.filter((cat) => !buckets[cat]?.length);
```
rugeli: nit: not a big deal for small arrays, but we're iterating `selectedCategories` twice which can be DRYed

something like:
```
const nonEmpty = []
const empty = []
for (const cat of selectedCategories) {
        (buckets[cat]?.length ? nonEmpty : empty).push(cat);
    }
...
```

